PR #8, opened 2026-07-03
feat: optional hooks around long e-ink BUSY waits

BrianPugh:
Equivalent of https://github.com/crosspoint-reader/community-sdk/pull/21; copying/pasting body below:

An e-ink refresh takes ~0.3–2 s, during which `pollBusy()` spins on the BUSY pin at whatever clock the host is running. This PR adds `setBusyWaitHooks(begin, end)`: plain function pointers fired when a wait exceeds 20 ms and again when it completes, letting host firmware apply its own power policy for the wait window (CrossPoint uses it to drop the ESP32-C3 to 10 MHz there).

- **Fully backwards compatible**: both hooks default to `nullptr`; behavior is unchanged unless installed. The only added cost is a null-check inside the 1 ms poll loop.
- The 20 ms threshold means short command waits never fire the hooks, so hosts don't pay two clock switches on trivial waits. The end hook fires only if the begin hook did.
- Covers both the X4 (active-HIGH) and X3 (active-LOW, two-phase) wait paths, including the X3 early-return path.

**Measured** on an X3 with a Nordic [PPK2](https://www.nordicsemi.com/Products/Development-hardware/Power-Profiler-Kit-2) at the battery terminals at 3.8v: with CrossPoint installing a 10 MHz downclock hook, a fast page turn drops from ~38 mC to ~29 mC (−24%), with fast/full/grayscale refreshes visually unchanged on device.

Supports https://github.com/crosspoint-reader/crosspoint-reader/pull/2525

### DIFF
--- a/libs/display/FreeInkDisplay/include/FreeInkDisplay.h
+++ b/libs/display/FreeInkDisplay/include/FreeInkDisplay.h
@@ -120,6 +120,13 @@ class FreeInkDisplay {
   // See EpdBus::setBusyWaitHooks for firing semantics.
   void setBusyWaitHooks(void (*beginHook)(), void (*endHook)()) { _bus.setBusyWaitHooks(beginHook, endHook); }
 
+  // Optional slice hook replacing the BUSY poll delay once a wait has proven
+  // long, so host firmware can sleep through the refresh instead of polling.
+  // See EpdBus::setBusyWaitSliceHook for the contract.
+  void setBusyWaitSliceHook(bool (*sliceHook)(int8_t busyPin, uint8_t busyLevel)) {
+    _bus.setBusyWaitSliceHook(sliceHook);
+  }
+
   // Access to frame buffer
   uint8_t* getFrameBuffer() const { return frameBuffer; }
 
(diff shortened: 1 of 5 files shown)


itsthisjustin:
If you've still got an x3 open and a way to scope it...

Can you put a logic analyzer on the panel SPI while stock firmware renders a sleep wallpaper. I need to know the actual CDI value, the exact plane bytes for a known image, and the timing. I want to diff our command stream against stock's byte-for-byte.

BrianPugh:
I have a cheap Salae-clone that I've never used but have been meaning to 😄 . In prep, from looking at the PCB/code, I *think* the following is the description of the testpads. I would still need to probe/find the following:

* DISPLAY_CS - GPIO21, Physical Pin 28
* D/C - GPIO4, Physical Pin 9
* BUSY - GPIO6, Physical Pin 12
* RST - GPIO5, Physical Pin 5

There's a few test points on the right-side of the PCB, but I bet they're unrelated. I also haven't yet lifted up the tape covering the display/pcb connector; there could be more test points over there.

20240604_004615 copy

PR #10, opened 2026-07-08
CrossPoint display compatibility + async X3 refresh + BUSY-edge ISR waiting

jpirnay:
This PR lands the display-layer changes needed to let Witch Reader firmware build against the FreeInk SDK via the `EInkDisplay` alias, and improves refresh latency/waiting for both X3 and X4 panels.

### What changed

- **WitchReader compatibility surface** ([`61aa2aa`](https://github.com/Free-Ink/freeink-sdk/commit/61aa2aa))
  - Added `displayStart()` / `displayFinish()` split to `PanelDriver`, defaulting to the old blocking `display()` so only X3 opts in.
  - `Uc8253X3Driver` splits `display()` at the `X3_DRF` wait so the waveform runs while the render task does non-SPI work.
  - Added shims to `FreeInkDisplay`: `triggerDisplay`/`completeDisplay`, `isRefreshPending`, `isX3Mode`, red-RAM sync helpers, `setSingleBufferFastDiff`, and `setFastGrayscaleLut`.
  - Made dual-buffer helpers available in every dual-buffer build, not just PSRAM, so CrossPoint's C3 can lend the secondary buffer out of DRAM.

- **20 MHz SPI for X4/X3** ([`9bd931e`](https://github.com/Free-Ink/freeink-sdk/commit/9bd931e))
  - Both the SSD1677 (X4) and UC8253 (X3) controllers are rated for 20 MHz; X4 was previously limited to 5 MHz and Witch Reader ran 40 MHz out of spec. Both profiles now use the datasheet maximum.

- **BUSY-edge ISR waiting** ([`7b7e15d`](https://github.com/Free-Ink/freeink-sdk/commit/7b7e15d), [`d9144ac`](https://github.com/Free-Ink/freeink-sdk/commit/d9144ac))
  - Replaced the 1 ms polling loop with a GPIO-interrupt + binary semaphore in `EpdBus::waitRefreshComplete()`.
  - The ISR path is now used for both X3 and X4; the function self-confirms the waveform is running before arming, so callers don't need a precondition.

- **X4 fast-DU refresh shortcut** ([`eaa73d5`](https://github.com/Free-Ink/freeink-sdk/commit/eaa73d5))
  - Added `fastDuRefreshShortcut(base)` so SSD1677 boards validated for true partial update can opt into the `CTRL2=0x1C` fast path (~77 ms saved per refresh on X4).
  - X4 opts in; other SSD1677 boards keep the conservative `0xFC` default until validated.

- **X4 overclocking option** ([`4003eaf`](https://github.com/Free-Ink/freeink-sdk/commit/4003eaf)) this will remove another 19ms due to faster SPI speed

### Files changed

- `libs/display/FreeInkDisplay/include/FreeInkDisplay.h`
- `libs/display/FreeInkDisplay/src/FreeInkDisplay.cpp`
- `libs/display/FreeInkDisplay/src/bus/EpdBus.cpp`
- `libs/display/FreeInkDisplay/src/bus/EpdBus.h`
- `libs/display/FreeInkDisplay/src/driver/PanelDriver.h`
- `libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp`
- `libs/display/FreeInkDisplay/src/driver/Uc8253X3Driver.cpp`
- `libs/display/FreeInkDisplay/src/driver/Uc8253X3Driver.h`
- `libs/hardware/BoardConfig/include/BoardConfig.h`


BrianPugh:
I'm playing around with this PR, and Fable had the following to say:

---

`waitRefreshComplete()` bypasses the **slice hook** (`setBusyWaitSliceHook`, added in c8ed8d5). The slice hook lives in `waitBusy()`'s `busyIdle()` poll loop, and this PR switches the two long BW waveform waits (X4 `refresh`, X3 `X3_DRF`) off that loop onto the semaphore wait — so for hosts that install a slice hook, it silently stops firing exactly where it matters most.

CrossPoint (in https://github.com/crosspoint-reader/crosspoint-reader/pull/2525) uses the slice hook to replace each poll delay with a light sleep that wakes on a GPIO **level** trigger when BUSY leaves its active state (level wakes work in light sleep). That has two consequences here:

1. **Power**: a task blocked on `xSemaphoreTake` leaves the chip in idle-task `waiti` at whatever clock the begin hook set — measurably more than actual light sleep. On an X3 measured with a PPK2 at the battery terminals, a fast page turn is ~26.5 mC with light-sleep slices vs ~29 mC with downclock-only waiting (which is what the semaphore wait is equivalent to). So merging as-is would regress hook-installing hosts by ~9% per refresh.
2. **Wake latency**: the slice hook already wakes on the exact BUSY transition (GPIO level wake), so the ISR path's precision advantage over the 1 ms poll doesn't apply to hosts using slices — they already have it, at lower power.

There's also a latent hazard: edge interrupts are *not* light-sleep wake sources, so if a host ever light-sleeps (from another task) while the edge ISR is armed, the completion edge is missed and the wait eats the full 30 s timeout. CrossPoint's power manager happens to guard against sleeping mid-render today, but the "host must not light-sleep while a refresh wait is armed" contract is currently implicit.

Both are addressed by having the ISR path defer to a host-installed slice hook:

```cpp
void EpdBus::waitRefreshComplete(const char* tag) {
  // A host slice hook (e.g. light sleep with GPIO level wake) beats a
  // blocked-task idle and already wakes on the exact BUSY transition —
  // prefer it. The ISR wait below serves hosts with no hooks installed.
  if (_busyWaitSliceHook) {
    waitBusy(tag);
    return;
  }
  ...
}
```

With that early-out, hosts with no hooks (Witch Reader) get the full benefit of the ISR wait, and hook-installing hosts keep their power policy. Happy to test the combined result on X3/X4 hardware with the PPK2 once updated.

jpirnay:
Thanks for the feedback, will introduce it, have found some more kinks, so putting this as draft for now

jpirnay:
Should be fine for review now - I still need to run official crosspoint-reader against it - until then I will leave it as draft. Witch(hunt) Reader does seem to work with it (x3 might still do one or two unnecessary half-refreshes every now and then compared to modified open-x4-sdk I have been using so far)

itsthisjustin:
@jpirnay I made a change to the x4 luts per OEM that conflict with this now. If you wanna test and give feedback, I'm not married to it but testing with users showed that it removed gray blotches from sleep screen and prevented ghosting

itsthisjustin:
@jpirnay did the same thing you were doing here for the x4. I did set 16mhz clock speed to be conservative...i imagine stock does 5mhz for SOME reason. If you see absolutely no reason not to do 20mhz, we can change it. Can we get this PR ready for merge? I think it solves a ton on both devices. Thanks so much for your work on this

itsthisjustin:
> @jpirnay did the same thing you were doing here for the x4. I did set 16mhz clock speed to be conservative...i imagine stock does 5mhz for SOME reason. If you see absolutely no reason not to do 20mhz, we can change it. Can we get this PR ready for merge? I think it solves a ton on both devices. Thanks so much for your work on this

I'll solve conflicts, so no worries on that

itsthisjustin:
@jpirnay consolidated all the work into this PR now. I wrote a migration guide for witch hunt here:

# Deferred refresh: one split interface

This change collapses the two parallel deferred-refresh mechanisms —
`PanelDriver::displayAsync()` (SSD1677/X4) and `displayStart()/displayFinish()`
(UC8253-X3) — into the single split interface, and merges the facade's two
pending flags into one. It builds on the `crosspoint-async-display-split` and
`x4-async-trigger-split` work; nothing conceptual is new, there is just one
mechanism now instead of two.

## Why

- `displayAsync` and `displayStart/displayFinish` expressed the same idea
  ("fire the waveform, come back later") with different plumbing, doubling the
  facade states (`_asyncPending` + `_splitPending`), the finish paths
  (`finishDisplayAsync` vs `completeDisplay` vs plain `waitBusy`), and the
  capability signals (`supportsAsyncDisplay()` vs displayStart's return value).
- The `waitBusy("async refresh")` drain for the X4 async path bypassed the ISR
  edge wait and, on drivers that fell back to blocking `display()`, spun the
  X3TwoPhase edge-detect timeout (1 s) against an already-idle panel.

## Driver interface (PanelDriver)

- **`displayAsync()` is removed.** The split is the one deferred mechanism:

  ```cpp
  // fire the waveform; true = deferred (displayFinish() must run), false = completed inline
  virtual bool displayStart(EpdBus&, const uint8_t* fb, const uint8_t* prev, RefreshMode, bool turnOff);
  // wait out the waveform + any post-waveform pipeline (X3: DTM1 sync + conditioning)
  virtual void displayFinish(EpdBus&, const uint8_t* fb);
  ```

  Defaults are unchanged (blocking `display()`, return false / no-op), so
  drivers without deferral need no edits.

- **`supportsAsyncDisplay()`** now means "displayStart() defers". It must agree
  with what `displayStart()` returns. It exists so the facade can skip shadow
  setup without a trial call, and so hosts can size overlap buffers up front.
  Overrides: SSD1677 → true, UC8253-X3 → true (new — this is what lets hosts
  run the same overlap flow on both panels).

- **SSD1677** gains the split: `displayStart()` is the old `displayAsync` body
  (displayImpl async path, returns true), `displayFinish()` is
  `bus.waitRefreshComplete("refresh")` (ISR edge wait; the done-level fast path
  handles a waveform that finished before the host got around to waiting).

## Facade (FreeInkDisplay)

- **One pending flag** (`_refreshPending`) replaces `_asyncPending` +
  `_splitPending`. It is set from `displayStart()`'s return value and drained
  in exactly one place (`syncPendingAsync()`), which always finishes through
  `driver->displayFinish()` — so the X3 post-waveform pipeline can never be
  skipped by a plain busy-wait again, and the X4 wait is always the ISR path.
- **Entry points are unchanged** — all of these still exist and now delegate to
  the same `displayAsyncImpl(mode, turnOff, noShadow)`:
  - `displayBufferAsync(mode)` — shadowed async (single-buffer allocates the
    48 KB shadow; caller may redraw fb immediately).
  - `displayBufferAsyncNoShadow(mode)` — no shadow; caller keeps fb intact
    until the wait and re-seeds the baseline itself (the tiled-AA flow).
  - `triggerDisplay(mode, off)` / `triggerDisplayAsync(mode, off)` — CrossPoint
    compat names, same semantics as before.
- **Finish/wait are aliases of one function**: `waitRefreshComplete()`,
  `completeDisplay()`, and `finishDisplayAsync()` all drain the single pending
  state. Call whichever your codebase already uses.
- `refreshBusy()` no longer clears the pending state when the pin reads idle —
  the driver's post-waveform work must still run. When it returns false, call
  any of the finish aliases (or any blocking display op, which self-heals).
- One behavior fix: single-buffer `displayBufferAsync()` on the **X3** now
  takes the blocking path instead of the shadow path. The shadow contract
  ("redraw fb immediately") is unsound on X3 because `displayFinish()` re-reads
  fb for the DTM1 sync; use `displayBufferAsyncNoShadow()` or
  `triggerDisplay()` (frame-intact contracts) for X3 overlap.

## What Witch Reader needs to change

Almost nothing at the call sites — the compat surface is intact. Checklist:

1. If any code overrides or calls `PanelDriver::displayAsync()` directly,
   port it to `displayStart()`/`displayFinish()` (see the SSD1677 diff for the
   1:1 mapping).
2. If any code assumed `refreshBusy() == false` implies "fully done", add a
   `waitRefreshComplete()` (cheap no-op when nothing is pending).
3. Custom deferring drivers must override `supportsAsyncDisplay()` to true.

## X4 fast-DU shortcut is now opt-in

`fastDuRefreshShortcut` (CTRL2=0x1C instead of the stock 0xFC partial
sequence) is no longer applied to the X4 by default — it is gated behind
`-DFREEINK_X4_FAST_DU_SHORTCUT`. The 0x1C path skips the per-refresh
temperature load and power sequencing; that is the community-sdk behavior the
stock-parity work moved away from after ghosting/blotching reports on some
panels, and its artifacts tend to appear only over long sessions and across
temperature. Worth ~85 ms/refresh where a panel is validated; enable the flag
per build once you have that validation, the mechanism is unchanged.

## What this buys

Measured on X4 (800x480, AA on, 16 MHz SPI): page turn 1274 ms → 822 ms by
overlapping the grayscale plane rendering with the BW waveform via the
no-shadow split. With X3 now reporting deferral through the same interface,
the identical host-side flow overlaps on X3 with no host changes.

libs/display/FreeInkDisplay/src/FreeInkDisplay.cpp
moben, line 578: Don't you have to `syncPendingAsync()` in all of these grayscale methods as they write BW / RED RAM?

I fear that might eat a lot of your gains.

itsthisjustin: The reader's tiled flow calls waitRefreshComplete() before its first plane write, which clears the pending state, so every guard added here evaluates to no-op in the flow that produces the numbers. No gains lost. 

Valid call out though and I did add a fix here

PR #23, opened 2026-07-28
feat: add abortable secure connection deadlines

rigogsilva:
## Summary

Add an optional, abort-aware connection API to `SecureClient` and thread `SecureHttpClient` request deadlines through TCP connection setup and TLS negotiation.

## Why

`SecureHttpClient::setTimeout()` previously bounded response processing but not the initial secure connection. `SecureClient` used a fresh hardcoded 15-second TLS deadline for automatic negotiation and another for its TLS 1.2 fallback. A caller with a shorter remaining operation budget could therefore remain blocked well beyond its advertised deadline.

## Changes

- Add `SecureClient::connectWithTimeout()` with one deadline shared by TCP setup, automatic TLS negotiation, and the TLS 1.2 fallback.
- Check an optional abort callback between TLS handshake iterations.
- Make `SecureHttpClient` pass its request timeout and tracked abort callback into secure connection setup.
- Bound plain HTTP TCP connection setup with the same request timeout.
- Preserve the existing `SecureClient::connect()` API and behavior for current callers.
- Preserve `SecureHttpClient::aborted()` classification when cancellation occurs during HTTP or HTTPS connection setup.

## Validation

- CrossPoint ESP32-C3 production firmware build: successful.
- CrossPoint cppcheck: no defects.
- CrossPoint host tests: 143/143 passed.
- Hardware boot validation on an Xteink X4.

## Context

This was developed for an opt-in CrossPoint KOReader automatic-progress upload flow that must finish or enter deep sleep within one bounded background-operation window. The API is general and leaves existing callers unchanged unless they opt into the deadline-aware connection method.

AI tools were used to assist implementation, review, and validation.

### DIFF
--- a/libs/network/SecureNet/include/SecureClient.h
+++ b/libs/network/SecureNet/include/SecureClient.h
@@ -22,10 +22,14 @@
 #include <Client.h>
 #include <WiFiClient.h>
 
+#include <functional>
+
 namespace freeink {
 
 class SecureClient : public Client {
  public:
+  using AbortCallback = std::function<bool()>;
+
   SecureClient() = default;
   ~SecureClient() override;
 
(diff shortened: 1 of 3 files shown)


rigogsilva:
Closing this draft to keep the CrossPoint automatic-upload change self-contained. The feature will retain its application-level Wi-Fi/request budget and document the underlying TLS timeout as best-effort instead of adding an SDK dependency.

PR #28, opened 2026-07-31
Display cleanups: blit fixes, vendor-doc audit, and UC8279 800×480 support

itsthisjustin:
### Image positioning + transparent blit

- drawImage no longer snaps to byte boundaries. The byte-aligned fast path is preserved; a new per-pixel path places images at exact columns (e.g. a logo centered at x=180 no longer shifts to 176). Source row stride is ceil(w/8), so non-multiple-of-8 widths are handled too. 
- New drawImageTransparent / blitImage(transparent): paints black pixels only (AND), leaving existing white content untouched. -
 
 ### Cleanups
 
 - EpdBus: fillPlane streams wide rows in chunks instead of silently truncating at 128 bytes; removed the now-unused writeMirroredPlane. 
 - It8951Driver: clamped the 4bpp row expansion to its static buffer so a wider injected geometry truncates instead of overrunning. 
 - Ed2208M5Driver: partial-window yEnd made inclusive, matching xEnd. 
 - Removed dead triggerRefresh/fillPlane helpers (UC8279d, Murphy) and the unused lut_factory_fast / lut_x4pro_partial SSD1677 tables.

### UC8179 

- GSST (0x65) now writes all 4 data bytes (was 2; the last two were left at reset defaults). 
- AA grayscale CDI now follows the first-vs-later rule: 0x29 07 on the first AA refresh after init, 0xA9 07 (border hold) on subsequent ones — stops the border being re-driven on every AA page. - Doc-correct register naming: 0x03 is PFS (was mislabeled PLL), 0xE0 is CCSET, 0xE5 is TSSET; config fields renamed to match (pfs, ccset, tsset, tssetFast, gateScan). Values unchanged.

### SSD1677 
- External-LUT AA activation now always sends 0x22 = 0xCC (previously 0x0C when the rails were already up) and marks power off afterward, per the reference. The enable bits are no-ops on already-powered rails, so behavior on the validated path is equivalent. 
- The default config now writes the stage-2 border (0x3C = 0xC0) explicitly with the LUT instead of relying on register carry-over.

### New: Uc8279X4Driver

- UC8279 800×480 panel variant Some X4-family units ship a UC8279 (VER LUT_VER byte 0x02/0x68; 0x69 reserved) instead of the UC8179. Previously these would have been routed to the UC8179 driver with the wrong init (2-byte CDI, wrong PSR, no gate offset).

### XteinkDetect 

- The probe now holds RST low 50 ms per the reference's identification path (was 1 ms; ~+100 ms boot across the two passes). 
- Promotion on SSD1677-default boards now decodes VER byte 2 (LUT_VER): 0x01 → UC8179, 0x02/0x68/0x69 → the new UC8279 driver, unknown → UC8179 fallback (benched units returning VER=00 00 01 FF FF route exactly as before).

### Logging fix

BoardConfig power-latch pin-collision warnings now use esp_rom_printf instead of Serial.printf, so they print before USB-CDC enumeration completes. Fixes cascade of warnings while compiling CrossPoint


itsthisjustin:
Confirmed no regressions on X4Pro with UC8179 controller or SSD1677. No regressions on x4 with SSD1677 either.

PR #35, opened 2026-08-13
fix: timeout issues

shirok1:
The timeout of TCP connection is incorrectly divided by 1000 (60s to 60ms), this will cause immediate timeout happen on certain special network environments e.g. tun2sock. Also timeout passed to wolfSSL is misconfigured to be a hard-coded 15s not the global timeout 60s.


itsthisjustin:
@shirok1 can you update this PR with a description of the issue and how you fixed it?

shirok1:
> @shirok1 can you update this PR with a description of the issue and how you fixed it?

Updated

PR #38, opened 2026-08-14
feat: opt-in fast-DU refresh shortcut for the Xteink X4 Pro

hooligan333:
## Summary

Adds an opt-in `FREEINK_X4PRO_FAST_DU_SHORTCUT` define that gives the Xteink X4 Pro the same differential fast-DU refresh shortcut the X4 already has behind `FREEINK_X4_FAST_DU_SHORTCUT`.

Today `Ssd1677Driver` pins `XteinkX4Pro` to the stock absolute fast sequence (`0xFC`), which reloads temperature and power-cycles the panel on every refresh (~500 ms measured on stock). The `0x1C` differential fast-DU path — already implemented and field-proven on the X4's SSD1677 glass — has no X4 Pro branch. This change adds one: under the define, `ssd1677X4ProConfig()` applies the existing `fastDuRefreshShortcut()` transform for `Board::XteinkX4Pro`, saving roughly ~85 ms per page-turn refresh.

## Gating

- Off by default; without the define the driver is byte-identical to today for every board.
- SSD1677-scoped only: newer X4 Pro batches that detect as UC8179/UC8279 never reach this config and are unaffected.

## Testing

Tested on X4 Pro hardware (SSD1677 batch) over multi-day reading sessions: page turns confirmed visibly faster by the device owner, with a ghosting/darkening soak across long sessions. One early ghosting report during testing was root-caused to an unrelated stale SDK pin in the test harness (missing grayscale fixes already on `main`), not to this shortcut — re-testing on current `main` was clean and the owner reports it working great.

Since fast-DU on this glass class has historically been panel-batch sensitive, keeping it opt-in until it accumulates wider soak coverage seems right; happy to follow up with a default-on change for `XteinkX4Pro` if/when you're comfortable.

https://claude.ai/code/session_01UUgQffv7GGVCWQ9gijsVbp

### DIFF
--- a/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
+++ b/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
@@ -135,6 +135,21 @@ static const Ssd1677Config& ssd1677X4Config() {
 }
 #endif
 
+// Xteink X4 Pro with the fast-DU shortcut — OPT-IN via
+// -DFREEINK_X4PRO_FAST_DU_SHORTCUT. The X4 Pro paints on the stock X4 config
+// (same GDEQ0426T82 panel class — see ssd1677ActiveConfig), so the same
+// ~85 ms/refresh win applies, and so does the same panel-variance caveat: 0x1C
+// skips the per-refresh temperature load and power sequencing, and artifacts
+// tend to appear only over long sessions and across temperature. Enable only
+// after validating on your unit. SSD1677-batch units only — UC8179/UC8279
+// batches select a different driver and never reach this config.
+#ifdef FREEINK_X4PRO_FAST_DU_SHORTCUT
+static const Ssd1677Config& ssd1677X4ProConfig() {
+  static const Ssd1677Config cfg = fastDuRefreshShortcut(ssd1677DefaultConfig());
+  return cfg;
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


itsthisjustin:
Any risk in just making this the default behavior?

hooligan333:
Some. The SDK originally defaulted this panel class to 0x1C, and the stock-parity work reverted to 0xFC after "heavy ghosting on some panels" reports (comment ssd1677DefaultConfig), which is why the X4's sibling flag is opt-in. My X4 Pro (SSD1677 batch) has been clean over long trials, but that doesn't necessarily speak to other panels... Though periodic full refreshes still reload temperature and scrub ghosting, UC-batch units never reach this config, and I think worst case would be ghosting between full refreshes. Perhaps worth testing in beta?

itsthisjustin:
> Some. The SDK originally defaulted this panel class to 0x1C, and the stock-parity work reverted to 0xFC after "heavy ghosting on some panels" reports (comment ssd1677DefaultConfig), which is why the X4's sibling flag is opt-in. My X4 Pro (SSD1677 batch) has been clean over long trials, but that doesn't necessarily speak to other panels... Though periodic full refreshes still reload temperature and scrub ghosting, UC-batch units never reach this config, and I think worst case would be ghosting between full refreshes. Perhaps worth testing in beta?

I'll take a look

PR #39, opened 2026-08-14
feat: light-sleep-surviving frontlight (FREEINK_FRONTLIGHT_LS)

hooligan333:
## Summary

Adds an opt-in `FREEINK_FRONTLIGHT_LS` define that makes the LEDC frontlight survive automatic light sleep, so firmwares using light sleep no longer have to choose between the frontlight and sleeping (today the practical workaround is "prevent light sleep while the frontlight is active", which forfeits the entire sleep saving for anyone reading with the light on — i.e. most night reading).

Two commits:

1. **`FREEINK_FRONTLIGHT_LS` + polled BUSY-wait grace.** Under the define, `FrontlightManager` attaches its channels through the IDF LEDC driver directly (the Arduino helpers don't expose `sleep_mode`): timer clocked from **RC_FAST** (~17.5 MHz, the LEDC source that keeps running through light sleep) with `LEDC_SLEEP_MODE_KEEP_ALIVE`. RC_FAST at 10 kHz supports 10-bit resolution, so the board profiles' full duty range — including `setBrightnessLevel`'s level-1 minimum step — stays expressible. Companion change in `EpdBus::waitRefreshComplete`'s slice-hook fallback: a 20 ms BUSY-assert grace loop, because a host sleeping in level-polled waits can miss the BUSY edge the ISR path relies on.
2. **Keep the RC_FAST keep-alive armed only while the light is actually lit.** The KEEP_ALIVE channel config takes a refcounted +1 on the RC_FAST sleep sub-mode at attach time — which, on the S3, keeps RC_FAST *and* the digital domain at its higher (~1.1 V) sleep bias powered through **every** light-sleep window from boot, even with the light off. `begin()` balances that boot-time +1 (via `esp_private/esp_sleep_internal.h` — the same refcounted API the LEDC driver itself manages this through; no public API exists) and `apply()` re-arms it only on 0↔nonzero total-duty transitions. Dark idle sleeps at full depth; the cost is paid only while lit. The refcount is provably balanced across single/dual-channel boards, partial attach failures, and repeated `begin()` calls.

## Gating

Off by default. Without the define, `attachChannel` keeps the existing Arduino paths (it now returns a success bool, otherwise identical) and no sleep sub-mode call is ever made. The EpdBus grace loop only runs in the slice-hook fallback path, which only engages when a host installs a busy-wait slice hook.

## Testing — on-device measurements (Xteink X4 Pro)

Validated on X4 Pro hardware under an `esp_pm`-based automatic-light-sleep firmware (`CONFIG_PM_ENABLE` + tickless idle, DFS 80–240 MHz), with this flag providing the frontlight path. Residency was measured with `CONFIG_PM_PROFILING` and `esp_pm_dump_locks()` after a ~31.5-minute session that included active reading:

```
Mode stats:
Mode      CPU_freq    Time(us)    Time(%)
SLEEP     80 M        1746410029  92%
APB_MIN   80 M        39793022    2 %
CPU_MAX   240M        106773643   5 %

Sleep stats:
light_sleep_counts:34492  light_sleep_reject_counts:0
```

Reading the numbers:

- **92 % of total uptime in light sleep** — and that includes time attached to a USB host with sleep suppressed, so residency during the detached reading window was ~95 %+.
- **34,492 sleep entries, 0 rejects** — no driver or lock ever refused a sleep attempt with this frontlight configuration active.
- **Average sleep window 1,746 s / 34,492 ≈ 50.6 ms** — exactly the firmware's 50 ms touch-poll cadence, confirming the wakes are the host's own timers, not LEDC/frontlight interference.
- The frontlight stays visibly lit and blink-free through the sleep windows (KEEP_ALIVE + RC_FAST doing their job), including at the lowest brightness levels, and the sub-1 % duty floor behaves as described above.

The device owner has run this in day-to-day reading (combined with other in-test features) and reports it working great.

https://claude.ai/code/session_01UUgQffv7GGVCWQ9gijsVbp

### DIFF
--- a/libs/display/FreeInkDisplay/src/bus/EpdBus.cpp
+++ b/libs/display/FreeInkDisplay/src/bus/EpdBus.cpp
@@ -312,6 +312,20 @@ void EpdBus::waitRefreshComplete(const char* tag) {
   // slice hook already delivers GPIO-precise wake, so the ISR path buys these hosts
   // nothing — fall back to the hooked poll.
   if (_busyWaitSliceHook != nullptr) {
+    // Refresh-completion context: BUSY assertion can trail MASTER_ACTIVATION
+    // by a few microseconds, and the ActiveHigh polled path (unlike ActiveLow's
+    // 100 ms grace loop, or the ISR path's 20 ms edge wait below) would fall
+    // through immediately — returning mid-waveform, after which single-buffer
+    // drivers rewrite controller RAM while the panel is still driving. Give
+    // the poll the same bounded grace here, in the refresh-only context, so
+    // waitBusy() itself (which also serves command waits where BUSY may never
+    // assert) stays untouched for every other caller.
+    if (_busy == BusyPolarity::ActiveHigh) {
+      const unsigned long graceStart = millis();
+      while (digitalRead(_pins.busy) != HIGH && millis() - graceStart < 20) {
+        delayMicroseconds(200);
+      }
+    }
     waitBusy(tag);
     return;
   }
--- a/libs/hardware/FrontlightManager/include/FrontlightManager.h
+++ b/libs/hardware/FrontlightManager/include/FrontlightManager.h
@@ -68,6 +68,17 @@ class FrontlightManager {
   // Recompute and write both channels from _brightness + _warmPercent.
   void apply();
 #endif
+#ifdef FREEINK_FRONTLIGHT_LS
+  // Keep RC_FAST powered through light sleep only while the light is actually
+  // lit. The LEDC driver's KEEP_ALIVE config pins RC_FAST (and the digital
+  // domain at its higher sleep bias) for every light-sleep window from boot;
+  // begin() cancels that via the refcounted sleep sub-mode API and apply()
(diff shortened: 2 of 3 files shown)


itsthisjustin:
Can you share the firmware side of this or open a pr for crosspoint so I can get this integrated prior to release this weekend?

hooligan333:
Firmware side is up: crosspoint-reader/crosspoint-reader#3032 (against `feat-x4-papermono-support`, per your request). It enables the define in the x4pro env, compiles out the frontlight-blocks-sleep exclusions in `HalPowerManager` under it, and pins the submodule to this PR's head — happy to rebase the pin onto the merge commit once this lands.

itsthisjustin:
> Firmware side is up: crosspoint-reader/crosspoint-reader#3032 (against `feat-x4-papermono-support`, per your request). It enables the define in the x4pro env, compiles out the frontlight-blocks-sleep exclusions in `HalPowerManager` under it, and pins the submodule to this PR's head — happy to rebase the pin onto the merge commit once this lands.

Merged into my branch. I'll test and bump anything else needed tonight

PR #40, opened 2026-08-15
feat: Add FreeInkDisplay functions to work with absolute grayscale luts

theZiz:
This is needed for using the absolute factory lut of the SS1677 on X4 without device specific information in crosspoint reader.

### Summary

First of all I changed the naming from "factory" to "absolute", because it is very X4 specific that we got those lut from the original firmware. In fact the whole point is using a direct rendering approach instead of BW + 2 differential steps. So the naming "absolute".

The method `supportsAbsoluteGrayscale` is needed to make sure to use this functionality only on einks having absolute lut defined, which is right now only the ssd1677 afaik. Right now I am checking whether the device is an X4 in crosspoint reader and this is not correct as newer X4 might have a different controller.

`FreeInkDisplay::displayAbsoluteGrayBuffer` is basically a copy of `FreeInkDisplay::displayGrayBuffer` but with calling `_driver->displayGrayAbsolute` in the end.


itsthisjustin:
Ha I just fixed this too and opened the same PR. Compare implementations for me? https://github.com/crosspoint-reader/crosspoint-reader/pull/3046

theZiz:
Ha, but I was faster :joy: 

However, I don't see your PR in the freeSDK-space. :thinking: 

In the crosspoint-reader-space your fix does more or less the same, but
* keeps calling it factory mode instead of absolute and
* doesn't handle the other corner cases I found.

So imho my implementation does a bit more of the needed extra mile. :)

theZiz:
Tbh I don't see where your change is in the freesdk-space. Don't we work wir PRs here? :thinking:

itsthisjustin:
> Tbh I don't see where your change is in the freesdk-space. Don't we work wir PRs here? :thinking: 

For critical paths I commit directly a lot of times. Feel free to shore up two prs for each repo and we can get them merged together

theZiz:
Okay, I will update my PR based on your changes. :)

theZiz:
@itsthisjustin I updated my PR.

I renamed your function and removed the mentioning of the SSD1677 as the functionality itself is display controller independent. If something figures out an absolute lut for the different display controllers (e.g. by extracting them from original firmwares), they could also benefit from this approach and it would work directly in Crosspoint Reader. :)

I also kept the displayAbsoluteGrayBuffer` "shortcut" so that the lut stays in the driver. Again: Another implementation could use it's own absolute lut here.

theZiz:
Thanks for the 👍, any more to do for merging this, @itsthisjustin ? 🙂

theZiz:
I got a Seeed reTerminal Sticky and can verify, that the SSD1677 absolute luts from the X4 also work for the Sticky. Therefor I activated the feature in the Sticky config.
Before:
image
After:
image

(Of course this also needs Crosspoint Reader changes, for which a PR is prepared. Just for verfification and motivation)

theZiz:
I ran into an interesting problem, quite rarely, but annoyingly, that the rendering failed "halfway", looking like this:

image

This happened more often with 24 Bit BMPs than with pre-dithered 2 Bit Bitmaps, but I guess this is more a random effect around timings. However, I was able to reproduce the bug best with this.

I dug deeper into zgredex original work, everything I do is based on (github.com/crosspoint-reader/crosspoint-reader/pull/1614) and figured out, that he did slightly different `CMD_DISPLAY_UPDATE_CTRL2` data, especially did he not do `ANALOG_OFF|CLOCK_OFF` and even added a comment about this, which I copied:

```
    // Keep rails on after the factory gray paint. Running ANALOG_OFF/CLOCK_OFF in
    // the same activation can disturb settled gray particles and reintroduce smear.
```
This sound exactly like my problem above. With this change my image always looks as expected:

image

Just for reference, the original image looks like this:

image

theZiz:
Closed as integrated to recent commits on main

PR #42, opened 2026-08-15
feat: add GT911 multi-touch support

uxjulia:
## Summary

- expose allocation-free, fixed-size GT911 contact snapshots and a capability check for multi-touch support
- recognize intentional two- to four-finger translation gestures, with contact matching and guards that reject pinches, rotations, ambiguous frames, and replacement contacts
- preserve existing single-touch behavior by suppressing tap, hold, and single-swipe classifiers for every multi-contact sequence
- keep FreeInkUI's normal input adapter single-contact only so applications explicitly opt in to multi-touch routing


itsthisjustin:
Insanely cool

PR #46, opened 2026-08-17
fix: also refreshing for "turnOff" activated after first display in ssd1677

theZiz:
See discussion in: https://github.com/crosspoint-reader/crosspoint-reader/pull/3009

The SSD1677 only does the initial refresh if `turnOff` is false, which is the fadingFix in the Crosspoint Reader domain. This does not make sense in my opinion. At least it fixes the problem in the link above.

### DIFF
--- a/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
+++ b/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
@@ -425,24 +425,22 @@ void Ssd1677Driver::displayImpl(EpdBus& bus, const uint8_t* fb, const uint8_t* p
   // the single-pass HALF (0xD7 "_updateFull"); it never runs the multi-flash OTP
   // full waveform (0xF7 is a dead fallback branch there). Promote to HALF where
   // the board has one; boards without (Sticky) promote to their vendor FULL.
-  if (!turnOff) {
-    if (_needsInitialFull) {
-      // First paint after boot/wake must not be a differential FAST: it only drives
-      // pixels that differ from the RED baseline, so it can't clear whatever is
-      // physically on the panel (the boot screen) and that ghosts through. But a HALF
-      // or FULL the caller already asked for is non-differential (BYPASS_RED) and
-      // clears the panel + seeds the baseline on its own — honor it and just consume
-      // the one-shot. Only upgrade a FAST request. This keeps the boot logo (a HALF)
-      // from paying an extra multi-inversion FULL-waveform flash on top of its own.
-      if (mode == RefreshMode::Fast) mode = RefreshMode::Full;
-      _needsInitialFull = false;
-      mode = (_cfg.halfSeqOverride != 0) ? RefreshMode::Half : RefreshMode::Full;
-    } else if (!_isScreenOn && _cfg.fullSeqOverride == 0) {
-      // X4-class cold start: panel asleep -> a (warmed) HALF full-clear. Override
-      // boards skip this — their fast sequence self-powers, so _isScreenOn is false
-      // every page and forcing HALF would make every page a slow full-waveform flash.
-      mode = RefreshMode::Half;
-    }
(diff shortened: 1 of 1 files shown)


thiagokokada (commented):
Confirmed that the original issue ("Sunlight Fade Fix" enabled on Xteink X4 with SSD1677) is fixed by applying this patch + https://github.com/crosspoint-reader/crosspoint-reader/commit/c6fe67d48d8b65f5dfdefae31f06d0bc9826f267.

thiagokokada (approved):
CC @itsthisjustin.

thiagokokada:
> Confirmed that the original issue ("Sunlight Fade Fix" enabled on Xteink X4 with SSD1677) is fixed by applying this patch + [crosspoint-reader/crosspoint-reader@c6fe67d](https://github.com/crosspoint-reader/crosspoint-reader/commit/c6fe67d48d8b65f5dfdefae31f06d0bc9826f267).

Actually, this seems to have fixed the boot issue but now when I am in the read page I can't switch pages (I am not sure if the pages are not switching or just the page is not rendering). I can open the book menu though, it is really strange and I have no idea why this is happening.

Reverting to 1.6.0rc works fine. I also tested commit https://github.com/crosspoint-reader/crosspoint-reader/commit/6a501bba544d9e80598811dbdf2134d1bcb1ebd2 (1.6.0rc) with this patch applied, same issue, so my conclusion is that this PR is the cause.

thiagokokada (changes requested):
Dismissing my approval since the issue reported above.

theZiz:
> > Confirmed that the original issue ("Sunlight Fade Fix" enabled on Xteink X4 with SSD1677) is fixed by applying this patch + [crosspoint-reader/crosspoint-reader@c6fe67d](https://github.com/crosspoint-reader/crosspoint-reader/commit/c6fe67d48d8b65f5dfdefae31f06d0bc9826f267).
> 
> Actually, this seems to have fixed the boot issue but now when I am in the read page I can't switch pages (I am not sure if the pages are not switching or just the page is not rendering). I can open the book menu though, it is really strange and I have no idea why this is happening.
> 
> Reverting to 1.6.0rc works fine. I also tested commit [crosspoint-reader/crosspoint-reader@6a501bb](https://github.com/crosspoint-reader/crosspoint-reader/commit/6a501bba544d9e80598811dbdf2134d1bcb1ebd2) (1.6.0rc) with this patch applied, same issue, so my conclusion is that this PR is the cause.

Hm, I can't reproduce what you are meaning. With "read page" you mean reading a book? Works as intended with my and your patch as well as the develop branch of Crosspoint Reader + my patch.

I can't really find any reason why my patch should make the thing worse at all. The logic is changed exactly one time for the very first display()-call. Afterwards it should not make any difference. I really don't get it.

Can you give me more clear productions steps?
* Which patch did you (not) apply?
* Does this happen with or without fading fix?
* With or without text anti aliasing?
* Are you entering the menu and then into text or do you boot directly into the book?

thiagokokada:
Sorry, just saw this right now.

>     * Which patch did you (not) apply?

I applied just this patch on top of the Free-Ink SDK that is set as a submodule in CrossPoint.

>     * Does this happen with or without fading fix?

I think I tested both.

>     * With or without text anti aliasing?

With.

>     * Are you entering the menu and then into text or do you boot directly into the book?

Menu -> Book.

theZiz:
Thanks for the answer though!

I can't think of ANY reason why this is happening with my change and think that something else is fishy (e.g. we had an USB freeze bug sometime in the past), but as this might also happen on other devices with other drivers, it is probably a good idea to have your change in the Crosspoint Reader instead.

I doubt that @itsthisjustin wants to take the risk for another change here while the problem is in fact already solved. Tbh I think the best would be now to remove the special handling code from here, because I think it has some logic flaws and other drivers don't have it, too. But on the other hand: Never touch a running system if not needed. :joy:

PR #47, opened 2026-08-17
LilyGo T5S3: profile corrections + LGFX grayscale fixes

jpirnay:
Found while bringing the LilyGo T5 E-Paper S3 Pro up on WitchReader. Two
independent commits, one file each.

## 1. `BoardConfig` — profile corrections (`2ebfc48`)

All four verified on hardware.

| Change | Why |
|---|---|
| `LILYGO_T5_PRO_GT911.hasHomeKey = true` | The board **has** a capacitive home key. The vendor wiki's button list (*"RST + BOOT + IO48 + PWR"*) omits it, so it was found by tracing the GT911 status bit `0x10`. `InputManager` reads that bit unconditionally, so detection always worked — the *consumers* gated on this flag (`wasHomeGesture()`, `wasHomeKeyHold()`) discarded every press. On a board whose only other nav input is one expander button, that is the difference between usable and not. |
| `sensors` → PCF8563 @ `0x51` | The board has a battery-backed RTC and the profile said `NO_SENSORS`, so it kept time in software and lost it on any real power cut. Schematic `hardware/T5 E-paper S3 Pro V1.0 24-12-24.pdf` page 3 / U3 shows a **PCF8563TS at 0x51**. The README's product table says PCF85063; the vendor's own `docs/pinmap.md` notes say to prefer the schematic and the mounted part. |
| `FREEINK_CAP_RTC` += `FREEINK_DEVICE_LILYGO` | Without it the `Rtc` lib links stub bodies and `begin()` returns false however correct the profile is. The two changes are only useful together. |
| `viewableInsets` sides `3 → 8` | This case sits closer over the glass than the X4's, whose value the default carries; 3px left the first and last characters of each line hard to read. Measured by eye in two passes; top/bottom are right at the defaults. Compare the X4 Pro's 7px sides. |

## 2. `LgfxEpdDriver` — grayscale path (`492c799`)

Both invisible until a host actually runs an AA pass on this panel.

**a. `displayGray()` treated its `fb` argument as the B/W base.** It isn't.
The facade passes `FreeInkDisplay::frameBuffer`, and the host's plane dance
(clear to `0x00`, render text-only, copy the plane out, call `displayGray`)
leaves the **last plane** there. Every background pixel therefore mapped to
`kGrayBlack`: the page inverted, the text vanished, and only the anti-aliased
marks remained.

`Ssd1677Driver::displayGray()` opens with `(void)fb` — its planes are already in
controller RAM and the panel retains the B/W image — so the argument was
effectively dead for the driver the caller was written against, and nobody
noticed it carried a plane.

Rather than push the contract back onto callers, the driver now overlays onto the
canvas the base push left behind: `pushSprite()` does not clear it, so the canvas
**is** the base. Only pixels a selector plane actually picks are darkened.

**b. `displayGray()` hardcoded `epd_fast`** while `display()` maps `HALF`/`FULL`
to `epd_text`. `Panel_EPD`'s per-pixel diff keys on the `epd_mode` LUT offset, so
a mode switch between base push and overlay re-drives every pixel — a full-screen
flash on any page refreshed with HALF or FULL. The mode of the last base push is
now remembered and reused, which is what the existing comment there already
claimed was happening.

## Validation

- Profile changes and **2a**: device-validated. The inversion is gone and
  anti-aliased text renders correctly.
- **2b**: reasoned from the driver's own documented behaviour and the arithmetic,
  **not** yet observed — the flash could not appear while the AA pass never
  reached the panel. Flagging that explicitly rather than implying it was tested.

### DIFF
--- a/libs/display/FreeInkDisplay/src/driver/LgfxEpdDriver.cpp
+++ b/libs/display/FreeInkDisplay/src/driver/LgfxEpdDriver.cpp
@@ -124,17 +124,6 @@ uint16_t g_w = 0, g_h = 0, g_wb = 0;
 
 constexpr uint8_t kGrayBlack = 0x00, kGrayDark = 0x55, kGrayLight = 0xAA, kGrayWhite = 0xFF;
 
-// (base, lsb, msb) per pixel -> 4 gray levels, matching the reference port: a set
-// base bit is white; otherwise msb/lsb pick light/dark; clear is black.
-inline uint8_t grayValue(uint8_t base, uint8_t lsb, uint8_t msb, uint8_t mask) {
-  if (base & mask) return kGrayWhite;
-  const bool l = lsb & mask, m = msb & mask;
-  if (m && l) return kGrayDark;
-  if (m) return kGrayLight;
-  if (l) return kGrayDark;
-  return kGrayBlack;
-}
-
 void allocCanvas(uint16_t w, uint16_t h) {
   g_w = w;
   g_h = h;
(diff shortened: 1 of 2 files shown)


itsthisjustin:
Thanks! This boards profile was severely lacking as I hadn't had time to improve upon it recently

PR #49, opened 2026-08-22
Add SSD1677 packed-window refresh pipeline and timing telemetry

integerQuant:
## Summary

- add explicit fast-refresh profiles and per-stage refresh timing telemetry
- add a bounded packed multi-window async API with safe unsupported-driver fallbacks
- extend the SSD1677 path with byte-aligned window transfer, one activation per batch, differential baseline replay, and Mode 2 state tracking
- keep existing full, half, and default fast-refresh behavior available

## Motivation

A low-latency terminal workload changes several small display regions per logical burst. The existing framebuffer API can redraw those cells, but cannot batch sparse panel-memory windows into one activation or expose enough timing data to distinguish transfer, waveform, and queue costs.

## Safety and scope

- packed-window refresh is bounded to 32 regions and validates coordinates, alignment, offsets, and byte counts
- unsupported drivers decline the operation without painting
- alternate SSD1677 terminal profiles are opt-in and restored by the owning activity
- no raw controller register or voltage interface is exposed
- this is intentionally a draft while the API shape and integration dependency are reviewed

## Validation

- built through the dependent CrossPoint `knietty_async_window` PlatformIO environment
- dependent native suite: 177/177 passed, including SSD1677 Mode 2 state coverage
- physically exercised on an XTEINK X4 through the dependent knietty firmware; window coherence, terminal operation, reader return, and sleep/wake gate passed

## Dependency

Consumer integration: crosspoint-reader/crosspoint-reader#3158.


integerQuant:
Closing this accidental upstream-targeted draft. The replacement fork-local draft is integerQuant/freeink-sdk#1.

PR #52, opened 2026-08-22
fix: recover CW2017 battery gauge updates

### LINKED ISSUE #44: On Xteink X4 Pro (CW2017), battery percentage remains the same until reboot
I haven't looked into this very deep, but at least on the CrossPoint end it calls readPercentageChecked(percent) successfully.

Daviex:
Summary

Fix the Xteink X4 Pro CW2017 battery percentage becoming stuck at one value until the device is put to sleep and woken again.

Fixes #44.

What was observed

On an Xteink X4 Pro running CrossPoint, the battery percentage could remain unchanged indefinitely while the device stayed awake.

CrossPoint continued calling BatteryMonitor::readPercentageChecked() successfully, but the returned SOC remained stale. Entering deep sleep and waking the device caused the value to update again.

The problem was traced to the CW2017 path in:

libs/hardware/BatteryMonitor/src/BatteryMonitor.cpp

Specifically, the initialization and recovery flow around:

* cw2017EnsureProfile()
* cw2017Reset()
* readGaugeSoc()

Root cause

The previous implementation treated CW2017 initialization as a one-shot operation:

static bool inited = false;
if (!inited) {
  cw2017EnsureProfile(g.gaugeAddr);
  inited = true;
}

This created several failure modes:

* cw2017EnsureProfile() returned void, so its caller could not determine whether initialization succeeded.
* inited was set to true even after an I2C failure, incomplete profile write, or readiness timeout.
* Profile verification treated a failed I2C read like a genuine profile mismatch, which could start rewriting BATINFO after a transient bus error.
* Return values from reset and profile-write operations were ignored.
* Once initialization was cached as successful, the driver did not revalidate the CW2017 mode, running version, or SOC value.
* A successful register transaction was therefore sufficient to keep returning a stale SOC, even if the gauge calculation engine was no longer in a healthy operating state.

A sleep/wake cycle restarts the application and runs the initialization path again, explaining why the percentage appeared to recover only after waking.

Fix

This change makes CW2017 initialization and runtime reads recoverable instead of permanently one-shot.

It:

* makes cw2017Reset() return success or failure and checks every MODE write;
* adds a bounded readiness wait for the gauge to leave its startup state and expose a valid SOC;
* separates an actual BATINFO mismatch from a profile-read I2C failure;
* verifies every BATINFO byte write and the update-flag write;
* makes cw2017EnsureProfile() return bool and propagate failures;
* retries failed initialization after a short bounded interval instead of caching the failure forever;
* validates MODE, VERSION, and SOC during sampled readings;
* clears the initialized state when the gauge is sleeping, starting, unavailable, or returning an invalid SOC, allowing the next call to run recovery;
* preserves the existing public BatteryMonitor API;
* leaves the BQ27220, ADC, and M5PM1 backends unchanged.

Hardware validation

Initial validation was performed on a physical Xteink X4 Pro with CrossPoint firmware built against this branch.

Before this patch, the displayed battery percentage remained fixed until a sleep/wake cycle.

After applying the patch, the percentage updates while the device remains awake, without requiring sleep or reboot.

Scope

Only the following file is modified:

libs/hardware/BatteryMonitor/src/BatteryMonitor.cpp

No public interfaces or board configurations are changed.

AI-assisted implementation disclosure

OpenAI Codex was used during this contribution to:

* inspect the FreeInk and CrossPoint battery-monitoring paths;
* identify the one-shot initialization and error-propagation defects;
* compare the CW2017 startup and profile-loading flow with reference driver implementations;
* draft and apply the C++ changes;
* review the resulting recovery logic;
* prepare this pull-request description.

The resulting patch was reviewed, committed to the contributor’s fork, built into the X4 Pro firmware, and tested on the physical device by the contributor.

The submitted source is ordinary C++ and has no runtime dependency on Codex or any other AI service.

### DIFF
--- a/libs/hardware/BatteryMonitor/src/BatteryMonitor.cpp
+++ b/libs/hardware/BatteryMonitor/src/BatteryMonitor.cpp
@@ -82,12 +82,18 @@ bool writeReg8(uint8_t addr, uint8_t reg, uint8_t val) {
 // XTEink::Cw2017PowerHal class in app1) via Ghidra. Unlike the BQ27220, the CW2017
 // reports 0% until a matching 80-byte BATINFO battery profile is resident, so init
 // must verify/upload one before SoC reads mean anything.
-constexpr uint8_t CW2017_REG_VERSION = 0x00;  // running state in (ver & 0xFD) == 0x0D
-constexpr uint8_t CW2017_REG_VCELL_H = 0x02;  // 14-bit VCELL, big-endian over 0x02/0x03
-constexpr uint8_t CW2017_REG_SOC = 0x04;      // integer percent (0x05 = fraction, unused)
-constexpr uint8_t CW2017_REG_MODE = 0x08;     // soft-reset / sleep control
-constexpr uint8_t CW2017_REG_CONFIG = 0x0B;   // bit7 = profile-loaded / update-enable
-constexpr uint8_t CW2017_REG_BATINFO = 0x10;  // 80-byte profile spans 0x10..0x5F
+constexpr uint8_t CW2017_REG_VERSION = 0x00;    // 0xA0 while starting; running versions match 0x0D/0x0F
+constexpr uint8_t CW2017_REG_VCELL_H = 0x02;    // 14-bit VCELL, big-endian over 0x02/0x03
+constexpr uint8_t CW2017_REG_SOC = 0x04;        // integer percent (0x05 = fraction, unused)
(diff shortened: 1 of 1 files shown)


itsthisjustin:
Thanks for doing this! Known bug, hadn't had time to look into it. Appreciate this PR so much

Daviex:
Thanks to you all for your work!

PR #59, opened 2026-08-27
X4 Pro: measured viewable insets, safeArea plumbed to DeviceContext

ma-r-s:
Three related changes from running CrossPlay on a physical Xteink X4 Pro:

1. **Measured viewable insets for the X4 Pro profile.** The bezel overlap was inherited from the X4 ('pending measurement' in the comment): top/bottom were eyeballed values and the 7px side inset was a scrollbar aesthetic, not a measurement. Measured on hardware with an on-screen 1px ruler: the glass hides 10 rows at the top, 1 column per side, 0 at the bottom (the panel sits shifted toward its bottom flex cable). Per-unit variance is real (crosspoint-reader#618 measured 5-11 hidden top rows across eleven X4 units), so units that still clip should be re-measured rather than padded; the comment now says so.

2. **GfxRendererTarget::deviceContext() fills DeviceContext.safeArea** from GfxRenderer::getOrientedViewableTRBL(), so Screen's content rect (seeded from safeRect()) keeps fui-laid-out screens clear of the covered pixels. Contexts built without the adapter are unchanged (safeArea stays zero).

3. **Screen::header(props, rect)** overload plus **Screen::setContentMarginAbsolute()**: the first lets chrome position the band itself (e.g. a band whose visible top tracks the safe area while the layout below keeps its own geometry) while sharing the anchor overload's theme substitution; the second is for hosts whose margins are already absolute-frame and insets-aware, where insetting from safeRect() would apply the safe area twice.

All three are shipping in CrossPlay (built for the X4 Pro) and were verified on the device; the firmware side that consumes them is in ma-r-s/crossplay.


prokrypt:
Just a note: x4 pro screen placement varies. On mine, a thin progress bar is invisible unless the device is tilted, and for another user it's fully visible with extra space on the bottom too.

prokrypt:
sticker_1787851608680
2 rectangles--one is 5 pixels from the edges, second is 10 pixels from the edges

itsthisjustin:
@ma-r-s can you check this against latest code in crosspoint for viewable insets and solve conflicts?

ma-r-s:
Merged latest main in, both conflicts resolved.

The insets conflict was just the new batteryChargeStatusActiveHigh field landing after I forked. Kept the measured values, added the field.

The margins one actually overlaps. setContentMarginFromScreen takes max(margin, safeArea) per side, setContentMarginAbsolute insets the full frame exactly, so passing zero gives you the whole frame instead of the safe area. CrossPlay uses that in about twenty places so I kept both. Say if you want mine gone.

PR #63, opened 2026-08-28
fix(x4pro): always park EPD panel before deep sleep

Belphemur:
PR crosspoint-reader/crosspoint-reader#3215 holds the X4 Pro master peripheral rail (power.latch0 = GPIO1) HIGH through deep sleep so battery-only sleep fast-wakes instead of cold-booting. Correct, but it exposed a latent driver bug: if the panel is not actually parked, the held rail keeps the panel's VCC up while the charge-pump/booster stays biased -> mA drain (the pre-PR state masked it because the latch floated LOW and the rail dropped).

The X4 Pro panel driver is boot-probed and may be promoted to SSD1677, UC8179 or UC8279X4 (XteinkDetect.cpp:334-353). Every driver gates the panel power-off on _isScreenOn while issuing DSLP unconditionally. Ssd1677Driver also desyncs the flag: after an AA/grayscale pass with turnOff==false the rails stay up (0xCC = CLOCK_ON|ANALOG_ON) but _isScreenOn is forced false (Ssd1677Driver.cpp:358), so powerOffController() early-returns (line 383) and DSLP is issued with the booster biased.

Make 'deepSleep() always parks the panel' an idempotent invariant:
- Ssd1677Driver: only clear _isScreenOn when turnOff actually powered down; drop the guard in powerOffController() so the booster-off always runs.
- Uc8179/Uc8279X4/Uc8279: drop the if(_isScreenOn) gate; always emit POF before DSLP (re-issuing POF on an already-off panel is harmless).

Builds clean for -e x4pro; pairs with crosspoint-reader#3215.

CF: https://github.com/crosspoint-reader/crosspoint-reader/commit/f0c655062c5135753134443d56513f8202208f6a#commitcomment-198022530

@itsthisjustin testing this on my side, don't hesitate to also test it.

Basically, we want to be sure the screen is off when we go to deepSleep

### DIFF
--- a/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
+++ b/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
@@ -389,7 +389,9 @@ void Ssd1677Driver::powerOn(EpdBus& bus) {
 void Ssd1677Driver::powerOffController(EpdBus& bus) {
   // Always park: re-issuing the border/analog-off sequence on an already-off
   // panel is harmless, and this guarantees the booster is off before DSLP even
-  // if _isScreenOn drifted out of sync (see the AA-pass desync above).
+  // if _isScreenOn drifted out of sync (see the AA-pass desync above). This
+  // driver is ActiveHigh (Ssd1677Driver.h), whose waitBusy() has a 30 s ceiling
+  // (EpdBus.cpp:225), so an off panel cannot hang here.
   bus.cmd(CMD_BORDER_WAVEFORM);
   bus.data(_cfg.borderWaveformInit);  // X4 Pro: 0x80
   bus.cmd(CMD_DISPLAY_UPDATE_CTRL2);
--- a/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
+++ b/libs/display/FreeInkDisplay/src/driver/Ssd1677Driver.cpp
@@ -354,8 +354,15 @@ void Ssd1677Driver::refresh(EpdBus& bus, RefreshMode mode, bool turnOff, bool as
     // not, so 0xCC is correct in both states. The production driver marks power OFF
     // after this pass; mirror that so the next refresh re-enables the rails.
     displayMode = 0xCC;
-    if (turnOff) displayMode |= 0x03;
-    _isScreenOn = false;
+    // Only mark the screen off when the activation actually powered it down. A
+    // custom-LUT/AA pass with turnOff==false leaves the rails ON (0xCC keeps
+    // CLOCK/ANALOG enabled); flipping the flag false here desyncs it from the
+    // real rail state and lets deepSleep() skip the booster-off sequence,
+    // leaving the charge-pump biased while PR #3215 holds the master rail up.
+    if (turnOff) {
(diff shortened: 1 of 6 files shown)


Belphemur:
## Re: "unconditional waitBusy() could hang forever in deepSleep()"

**TL;DR:** The mechanism is real (UcIdleHigh `waitBusy()` has no timeout at `EpdBus.cpp:251-267`), but the hang is **not reachable on the X4 Pro**, because the panel rail is *always powered* there — so BUSY is never stuck LOW at `deepSleep()`. That said, the concern is legitimate enough that I've pushed a hardening commit (`dd79e58`) that bounds the wait *without* masking real faults, and documents the invariant inline.

### Why it can't hang here
- The X4 Pro profile sets `display.powerEnable = PIN_UNASSIGNED` (`BoardConfig.h:1413`). `powerDownRailsForSleep()` (`PowerManager.cpp:82-83`) keys its reset-sleep level off that pin, so on the X4 Pro **the panel rail is never gated** — the controller stays powered through sleep.
- A powered UC81xx in idle presents BUSY **HIGH** (that's the definition of `UcIdleHigh`, `EpdBus.h:20`). So when `deepSleep()` issues `CMD_POWER_OFF`, the panel responds and `waitBusy()` returns immediately. There is no stuck-LOW state to loop on.
- The stuck-LOW scenario the review imagines — rail dropped, BUSY floats LOW — is exactly the **pre-#3215** battery-only bug (the latch floated LOW when external power left). PR #3215 fixes that by holding `power.latch0 = GPIO1` HIGH through deep sleep. My change operates *inside* that now-always-powered state, so the failure mode can't occur.

### Why a bounded-timeout wait was the wrong fix
The `UcIdleHigh` path deliberately omits a timeout (`EpdBus.cpp:254-256`): *"issuing the next command while BUSY_N is still LOW can make the UC controller discard plane or LUT writes."* A timeout that lets `deepSleep()` proceed to `CMD_DEEP_SLEEP` on an incomplete POF would trade a (non-existent here) hang for silently corrupting the power-down — and would *hide* a genuine hardware fault instead of surfacing it. Wrong trade.

### What I changed instead (`dd79e58`)
- **Keep** the unconditional `CMD_POWER_OFF` (idempotent — re-issuing POF on an already-off panel is harmless, and it's what guarantees the booster is off before DSLP).
- **Gate the `waitBusy()` on `_isScreenOn`**: only an *on* panel has real work to wait for, and on the X4 Pro that wait is instantaneous. An already-off panel skips the unbounded poll entirely — removing the only theoretical hang while preserving correctness.
- **Documented** the invariant in all three UC drivers and in `Ssd1677Driver::powerOffController` (which is `ActiveHigh` → already 30 s-bounded at `EpdBus.cpp:225`, so an off panel there can't hang either).

Builds green for `-e x4pro`. The "always park before DSLP" fix from the PR stands; this commit only tightens the power-down wait.

CC @itsthisjustin — still good to test on hardware (battery in series on a battery-only sleep: expect tens of µA; anything mA means the panel didn't park, which this change would not have caused).

Belphemur:
@itsthisjustin I think I got it, this is the first part of the fixes and then I'll have another PR for crosspoint to use the new front light methods

PR #65, opened 2026-08-29
fix: real peer verification in SecureClient (hostname check + fail-closed CA load)

sfoulad:
## Summary

`SecureClient::connect()` loads the caller's CA bundle via `setCACert()` but never calls `wolfSSL_check_domain_name()`, so **hostname verification is silently skipped**: any certificate signed by any anchor in the bundle is accepted for any host. A network attacker holding a valid certificate for *their own* domain (issued by the same public CA) passes verification when impersonating the target. This defeats the point of `setCACert` for public-internet endpoints.

Two further hardening items in the same small surface:

1. `wolfSSL_CTX_load_verify_buffer()` failure was not treated as fatal — a corrupt/truncated bundle degraded to an unverified connection. It now fails the connect (fail closed).
2. `SecureHttpClient` treated every response as having a body; per RFC 9110, 1xx/204/304 have none. A kept-alive `304 Not Modified` (conditional GET) previously stalled the read loop until timeout.

## Changes

- `SecureClient.cpp`: arm `wolfSSL_check_domain_name(ssl, host)` for every non-insecure connection; fail closed on CA-load errors. (+24/−2)
- `SecureHttpClient.h`: no-body status handling for 1xx/204/304. (+11)

`setInsecure()` behaviour is unchanged — callers that explicitly opt out still can; this only makes the *verified* path actually verify.

## Verification

- Builds clean against a CrossPoint-derived firmware on ESP32-C3 (wolfSSL path, X25519 + SP math config).
- Host-level chain fixtures: correct chain + correct hostname verifies; wrong CA fails; hostname mismatch fails (including a valid leaf for a different host).
- The RFC 9110 change is exercised by a conditional-GET 304 over a kept-alive connection.

Happy to split the RFC 9110 hunk into its own PR if you prefer reviewing the security change in isolation.

### DIFF
--- a/libs/network/SecureNet/include/SecureHttpClient.h
+++ b/libs/network/SecureNet/include/SecureHttpClient.h
@@ -262,6 +262,17 @@ class SecureHttpClient {
         return -1;
       }
 
+      // RFC 9110 §6.4.1: 1xx, 204 and 304 responses never carry a body,
+      // regardless of what Content-Length or Transfer-Encoding they echo (a
+      // 304 replays the cached response's headers). Without this, a 304 with
+      // no framing headers would fall into readUntilClose() and stall a
+      // kept-alive connection until the timeout.
+      if (_status == 204 || _status == 304 || (_status >= 100 && _status < 200)) {
+        _bodyComplete = true;
+        if (!_reuse || !keepAlive) closeConnection();
+        return _status;
+      }
+
       // A close-delimited body (no framing) ends WITH the connection, so it can
       // never leave a reusable socket behind.
       // A 3xx body that is about to be followed is protocol plumbing, not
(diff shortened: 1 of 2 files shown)


sfoulad:
Status update: this makes SecureClient actually verify the peer — hostname check plus fail-closed CA loading — instead of accepting any CA-signed certificate for any host; it's the SDK half of crosspoint-reader#3298. This repo runs no CI, so the evidence is downstream: it builds clean into an ESP32-C3 firmware and the chain/hostname negative tests pass at the HTTP layer against captured live GitHub chains. Downstream we're pinned to a fork commit for this one change; a review and merge lets our submodule return to an official upstream ref. Happy to split the hostname check and CA-load changes if that's easier to review.

itsthisjustin:
We kinda don't do this on purpose for memory purposes as well as the unlocker tool

PR #66, opened 2026-08-30
fix(x4pro): restore frontlight — 10kHz/10-bit RC_FAST-compatible PWM

Belphemur:
## Summary

The X4 Pro frontlight stopped working (stays dark) after upstream commit `13d23ba`
("Update X4 Pro display driver"), which bumped the X4 Pro frontlight to `25 kHz / 10-bit`.

Under `FREEINK_FRONTLIGHT_LS`, `FrontlightManager::begin()` clocks the LEDC timer from
**RC_FAST** (~17.5 MHz) so the PWM survives light sleep. At 10-bit resolution the fractional
divisor for 25 kHz makes `ledc_timer_config` return `ESP_ERR_INVALID_ARG` — both channels
(GPIO8 cool / GPIO9 warm) are left **unconfigured** and the light never comes on.

This reverts the X4 Pro profile to the working **10 kHz / 10-bit** (div ~438), restoring the
frontlight while keeping the full 10-bit dim floor.

## Scope / context

- This is the *remaining* frontlight fix for the X4 Pro. The wake-regression half
  (`park()` / `releaseOnWake()` unconditional hold release, SDK PR #63) is already merged into
  `Free-Ink/freeink-sdk` main. With that merged, the light wakes correctly but the *attach*
  step still fails at 25 kHz, so the panel stays dark — this commit closes that gap.
- One-line change in `BoardConfig.h` only; no driver/consumer code touched.
- The RC_FAST-clocked `FREEINK_FRONTLIGHT_LS` path is only meaningful for light sleep, which the
  CrossPoint firmware does not enter (`setPowerSaving` scales CPU frequency only), so 10 kHz is
  the safe, zero-risk choice here. If light-sleep LEDC support is ever dropped in favor of an
  XTAL/AUTO clock, this can return to the OEM 25 kHz.

## Verification

- Device: X4 Pro frontlight toggles on/off and ramps brightness after deep sleep (serial trace
  shows `apply: ... lit=1`).
- `ledc_timer_config` no longer returns `ESP_ERR_INVALID_ARG` for the X4 Pro frontlight channels.

Ported from the Belphemur fork (CrossPoint X4 Pro bring-up).

### DIFF
--- a/libs/hardware/BoardConfig/include/BoardConfig.h
+++ b/libs/hardware/BoardConfig/include/BoardConfig.h
@@ -1561,13 +1561,17 @@ constexpr BoardProfile XTEINK_X4_PRO = {
     // Frontlight: dual warm/cold LEDC PWM with color temperature (NVS lightWarmValue/
     // lightColdValue/lightCT/lightBri/lightOn). Recovered from the OEM LEDC init (IROM
     // 0x420a2130 → helper 0x420a20c0): two channels — GPIO8 on LEDC ch4 and GPIO9 on ch5 —
-    // The original bring-up dump used 10 kHz; stock 7.0.8 passes 25 kHz / 10-bit to
-    // the frontlight initializer on the same pins. Use that directly recovered value.
-    // Both channels are active-HIGH (init drives the pin LOW = off, brightness raises
-    // duty).
+    // The X4 Pro frontlight runs through the FREEINK_FRONTLIGHT_LS path, which clocks the
+    // LEDC timer from RC_FAST (~17.5 MHz) so the PWM survives light sleep. RC_FAST can only
+    // reach ~17 kHz at 10-bit resolution (divisor must exceed 256; 25 kHz/10-bit yields div
+    // 175 -> ledc_timer_config fails -> both channels unconfigured -> light stays dark). Use
+    // 10 kHz / 10-bit (the pre-regression value that worked): div ~438, valid, full dim floor.
+    // Both channels are active-HIGH (init drives the pin LOW = off, brightness raises duty).
     // GPIO8 is the hardware-confirmed cool channel and GPIO9 the warm channel;
     // FrontlightManager mixes them for color-temperature control.
-    {8, 25000, 10, true, 9},
+    // NOTE: if light-sleep LEDC support is ever dropped (driver uses XTAL/AUTO clock) this
+    // can safely return to the OEM 25 kHz.
+    {8, 10000, 10, true, 9},
     NO_AUDIO,
     NO_LEDS,
     NO_FLIP,  // panel mount transform pending hardware; native SSD1677 scan is 800x480 landscape


Belphemur:
@itsthisjustin fully tested to be working, still working on the battery drain on my fork, I'll upstream when I have a final fix, it's really hard to pinpoint it ...

itsthisjustin:
Interesting. Stock uses 25 and I ran this all night with it working fine

Belphemur:
That weird, I wonder if it can be related to other fixes I'm trying to do relating to battery drain and the two conflicted

Belphemur:
Yep, it relates to other commits I have in the firmware for parking and releasingOnWake the front light, weirdly enough for those action to work I need those value for the PWM ... I need to dig deeper what is the reason

Belphemur:
Do you know how I can get my hand on the stock firmware code ? Or even just the ota package if providing that is ... Problematic.

-------- Original Message --------

itsthisjustin:
It's available for download on the crosspoint site on the /debug page

Belphemur:
## TL;DR

**CrossPoint clocks the X4 Pro frontlight from `RC_FAST`; the stock OEM firmware does _not_.** That single difference is the entire reason the 25 kHz profile fails and we had to revert to 10 kHz in this PR.

- CrossPoint (this firmware, under `FREEINK_FRONTLIGHT_LS`): frontlight LEDC timer → **`LEDC_USE_RC_FAST_CLK`** (~17.5 MHz). At 10-bit, RC_FAST caps at ~17 kHz, so 25 kHz is unreachable → `ledc_timer_config` fails → light stays dark → this PR restores 10 kHz.
- Stock OEM (`xteink_app` v7.4.4): frontlight LEDC timer → **`LEDC_AUTO_CLK` → XTAL/APB** (~40 MHz). At 10-bit that ceiling is ~39 kHz, so stock's 25 kHz is valid and the light works.
- Stock also **deep-sleeps only** (no light sleep at all), so the "keep PWM alive through light sleep" rationale for RC_FAST doesn't even apply to stock. Add to that: stock never holds the master rail through sleep, so it has no frontlight leak to begin with.

So "stock runs 25 kHz fine" is true — but stock reaches it on XTAL, **not** on RC_FAST. The fork is what introduced RC_FAST into the frontlight path.

---

## Full analysis

### How we know this (no guesswork)
I downloaded the stock OTA from the CrossPoint site (`/debug` / flash-tools API → `xteink_app_update_x4pro_7.4.4_20260827_133901.xota`) and decompiled the decrypted ESP-IDF image (`xteink_app` v7.4.4, IDF v6.0.1, ESP32-S3) with the Espressif Xtensa objdump. This is reverse engineering for interoperability/debug — the binary is the vendor's own published firmware.

### 1. The frontlight frequency is real, but it's on XTAL
The frontlight channel config table in the stock binary has `25000` / `10000` at the duty/freq slots — confirming the OEM is **25 kHz / 10-bit**. But the *clock* that drives that timer is what matters:

- Stock frontlight init function: `0x4200F570` (calls `ledc_timer_config` at `0x4200e984`).
- It builds a `ledc_timer_config_t` on the stack whose `clk_cfg` field is **left at `0` = `LEDC_AUTO_CLK`** (struct zeroed in the prologue; no `movi aN, <clk>` stores a clock source before the call).
- IDF `ledc_set_timer_div` (`components/esp_driver_ledc/src/ledc.c:597`) with `AUTO_CLK` calls `ledc_auto_clk_divisor` (`ledc.c:562`), which **iterates candidate clocks and skips any whose divisor can't reach the requested frequency**.
- At 25 kHz / 10-bit:
  - RC_FAST: `17.5e6 / 25000 / 1024 ≈ 0.68` → divisor < 256 → **invalid** → AUTO skips it.
  - XTAL (40 MHz): `40e6 / 25000 / 1024 ≈ 1.56` → valid → AUTO selects it.
- Result: stock's 25 kHz frontlight runs on **XTAL/APB, not RC_FAST**.

Stock does contain ~13,894 RC_FAST clock constants and the `ESP_SLEEP_DIG_USE_RC_FAST_MODE` string — but those are for *other* sleep keep-alive paths, not the frontlight PWM.

### 2. Where RC_FAST in the fork actually comes from
CrossPoint's frontlight takes the RC_FAST path only because of `FREEINK_FRONTLIGHT_LS`, which was **added upstream** (not by us) in commit `cb7ec59` "Add light-sleep support" (Erica Jensen, PR #39). It pins `LEDC_USE_RC_FAST_CLK` so the PWM survives **light sleep**.

The problem: this firmware never enters light sleep — `setPowerSaving` only scales CPU frequency, and stock itself deep-sleeps only. So the RC_FAST choice buys us nothing here, while it **caps the frontlight at ~17 kHz at 10-bit**, which is exactly why 25 kHz can't be configured and we reverted to 10 kHz in this PR.

### 3. This is unrelated to the `park()` / `releaseOnWake()` leak fix
Worth keeping separate, because the two got tangled in the earlier debugging:
- The 10 kHz revert (this PR) is caused **solely** by RC_FAST's frequency ceiling.
- The deep-sleep frontlight leak is caused by the fork holding the master rail (GPIO1) HIGH for fast-wake (PR #3215), which exposed the frontlight driver. `FrontlightManager::park()` / `releaseOnWake()` fix it by driving GPIO8/9 LOW + `gpio_hold_en` — and those are **clock-agnostic**. They don't touch `clk_cfg` or the frequency.

So dropping RC_FAST does **not** reopen the leak.

### 4. The clean option (flagging for whoever merges)
If we drop `FREEINK_FRONTLIGHT_LS` / the RC_FAST pin for x4pro and let the timer fall back to `AUTO` (XTAL/APB), the frontlight can return to the OEM **25 kHz / 10-bit** — matching stock exactly — and `park()`/`releaseOnWake()` still close the master-rail leak. That's the path I'd upstream once the battery-drain work on my fork is finalized. This PR's 10 kHz is the safe stop-gap in the meantime.

### Bottom line for the thread
"Stock uses 25 kHz and works" — yes, because **stock is on XTAL/AUTO, not RC_FAST**. The fork is the one on RC_FAST, and that's the root cause of the 25 kHz attach failure this PR reverts. The battery-drain commits and the PWM clock are independent issues; conflating them is what made the earlier debugging loop.

Belphemur:
## Correction / clarification on the "stock uses XTAL not RC_FAST" claim

I need to walk back one part of my earlier comment. I stated as fact that stock clocks the frontlight from `LEDC_AUTO_CLK → XTAL` rather than RC_FAST. **That is not actually proven by the binary — I overstated it.**

What happened: I traced the wrong function as "frontlight init" and read a `clk_cfg = 0` from it. When I re-checked against the *real* `ledc_timer_config` (body at `0x4222446C`, identified by its `__func__` string), the stock frontlight-config module at `0x422e6756` calls it through a **register-indirect `callx8`** (C++ indirect call), so the `clk_cfg` value is passed in a register — not as a visible immediate. A flat `objdump` disassembly can't resolve that without a control-flow disassembler (Ghidra). An independent OpenCode static scan reached the same "needs Ghidra" conclusion.

**What IS solid from the stock binary (not changed):**
- Stock frontlight = **25 kHz / 10-bit** is real — the config table at load `0x3fc9ec9c` holds `25000`/`10000`, consumed by the `0x422e6756` frontlight module. Raw bytes, verified.
- RC_FAST at 10-bit caps at ~17 kHz, so **25 kHz is mathematically unreachable on RC_FAST** (pure divisor math + IDF `ledc.c`). This holds for any firmware.
- Therefore the fork's 10 kHz revert (this PR) is caused by `FREEINK_FRONTLIGHT_LS` pinning `LEDC_USE_RC_FAST_CLK` — that part stands. But whether stock's 25 kHz is on XTAL/AUTO or some other path is **open** until Ghidra reads the actual `clk_cfg` at the stock call site.
- Stock does have a light-sleep path (`esp_light_sleep_start` is linked, "Light Sleep" UI present), so RC_FAST keep-alive is at least *plausibly* used by stock somewhere — unlike the fork, which only deep-sleeps.

So: the headline of the earlier comment — "fork is on RC_FAST, stock is not" — should be read as "**fork is provably on RC_FAST (that's why 10 kHz); stock's 25 kHz clock source is not yet determined**." The safe, verified claim is only: stock reaches 25 kHz (something RC_FAST can't do), so it is *not* clocked purely from RC_FAST the way the fork is. If someone runs Ghidra and confirms stock passes `AUTO`/`XTAL`, then option B (drop RC_FAST for x4pro, keep `park()`) is the clean way back to OEM 25 kHz. Until then, treat that as a hypothesis.

Apologies for the over-confident wording in the first comment — recording the correction here so the thread isn't misleading.

itsthisjustin:
Fyi I also removed that light sleep for frontlight thing in the build I'm running

Belphemur:
In that case yeah this PR isn't needed since the math works since it disable the rc_fast on the front light. Is it part of reducing the possible light flickering that was reported on low battery?

-------- Original Message --------

itsthisjustin:
That was the idea at least yeah

Belphemur:
Closing this then, not needed when going back to stock and removing light sleep

Belphemur:
Also check on my build and reverting to stock and deleting LS works perfectly, going to see if the battery drain is also resolved with disabling the light fully when going to sleep.
I'm literally writing a csv file with all the mA values before and after sleep and mV too. Everything I can to get a good baseline (only in debug mode)

PR #69, opened 2026-08-31
fix(keyboard): give ף its own key and draw the language key as a globe

winst0niuss:
Follow-up to #29, acting on @Uri-Tauber's review of the firmware companion, [crosspoint-reader#2858](https://github.com/crosspoint-reader/crosspoint-reader/pull/2858#issuecomment-5478925684):

> a. I think its very confusing to have ף as long press on פ. Would make more sense to remove ' from the left and move the entire line one step left, leaving space for ף.
> b. I would prefer if the **change lang** key would have globe icon, instead of **ENG**/**HEB**.

## A key of its own for ף

`HE_ROW1` drops the geresh from its left end and shifts one step left; the freed slot at the right goes to ף. Still ten keys. #29 justified the long-press by the grid having no room, but that held only while the geresh took a slot, and the geresh is the cheaper thing to move: `'` is one tap away on the symbols layer, whereas a final form reached by holding its base letter is invisible unless you go looking.

## The language key becomes a glyph

`KeyKind::Lang` draws Lucide's globe, sharing the delete key's draw path. **This removes `KeyboardProps::langLabel`** and the `"EN"` placeholder in the built-in tables; a consumer that set the field just stops setting it. The label named the layout in use — something the letter keys already say, and in a three-letter code that needed translating besides.

The art is 32px, which is where the key sizer lands for a keyboard row — a disc with two meridians and an equator has no detail to spare at half that.

| QWERTY | Hebrew |
|---|---|
| QWERTY | Hebrew |

## Bottom row spacing

The language key takes **one unit rather than two**, giving the width to the space bar — a 28px glyph does not need 90px of key. The row still totals ten units, so `?123` and `OK` do not move. The **space rule now spans 90% of its key** rather than 60%: keys draw no background, so the eye measures gaps between glyphs, and a short rule in the widest key of the row read as a hole.

Against a letter row's own rhythm (Shift→z is 29.5px on a 480px panel), the three gaps go 50.5/67.5/63.5 → 27.7/19.9/38.6. They cannot be made equal while `?123` and `OK` hold their positions, since `OK`'s glyph sits 27px inside its own key; the ratio straddles the target instead.

The rule change reaches `EN_ROW4` and `SYMBOL_ROW4`, so single-script keyboards get a longer space bar too — the one visual change here not confined to multi-script layouts. `keyGrid` keeps 60%: its cells are uniform width, where a 90% rule would run into its neighbours.

## Testing

Exercised in the desktop simulator on Hebrew, Cyrillic and plain QWERTY: the globe in normal and selected state, ף and פ producing distinct code points, the geresh still on the symbols layer, and the bottom row against the rows above it. Host suite passes (2796 checks, 0 failed); the consuming firmware builds for ESP32-C3 with no `cppcheck` defects. Checked by script that every row's declared `count` still matches its key count and that the Hebrew layer has no duplicate ids.


Uri-Tauber:
@itsthisjustin Any thoughts?

itsthisjustin:
> @itsthisjustin Any thoughts?

Looks great to me

PR #70, opened 2026-09-01
feat(power): improve X4 Pro battery life - drop RC_FAST frontlight retention + add shutdown-reason RTC marker

Belphemur:
Prepare the FreeInk SDK for the crosspoint firmware changes that improve X4 Pro battery life.

This PR includes the SDK-side changes needed for the crosspoint battery-drain improvement:
- Drop RC_FAST frontlight retention in light sleep (removes LEDC_SLEEP_MODE_KEEP_ALIVE / FREEINK_FRONTLIGHT_LS)
- Restore 25 kHz frontlight PWM timer (was downgraded to 10 kHz by the RC_FAST workaround)
- Add shutdown-reason RTC marker so crosspoint can detect power-off shutdowns via RTC
- Add park() support for frontlight power-gating during sleep

The light-sleep RC_FAST keep-alive is intentionally removed per discussion with the @itsthisjustin  - upstream PR #39 compatibility is not preserved.

Crosspoint consumer-side changes (auto-shutdown timer, HalPowerManager, HalFrontlight) will follow in a separate draft PR after this merges.


Belphemur:
@itsthisjustin relates to what we talked about about disabling RC_FAST to keep the same refresh rate on the frontlight + other findings relating to how the stock firmware works.

I plan to add to upstream an auto shutdown feature, different from sleep, akin to what the stock firmware has, and store in the slow mem the reason (as stock does it) for the sleep.

PR #71, opened 2026-09-01
feat: support inline list section headings

oreglio:
Adds an optional inline section heading to ListItem while keeping the item as the only logical list index.\n\nThe heading and selectable row are fitted atomically, so a heading cannot be orphaned at the bottom of the viewport. The existing virtual-window and ListNav APIs remain the source of truth for paging.\n\nIncludes host coverage for virtual-window headings and the no-orphan boundary.\n\nThis is the SDK dependency for crosspoint-reader/crosspoint-reader#2888. The branch intentionally starts from the SDK commit currently pinned by that feature branch; this PR is draft pending coordination of the final SDK bump.


itsthisjustin:
Is this ready?

oreglio:
Sorry yes I forgot to make it ready

PR #82, opened 2026-09-06
fix(usb-msc): prevent crash on USB Drive eject/unplug

Belphemur:
## BLUF

USB Drive mode crashes (silent reboot, no crash log) when the host ejects or
unplugs during file transfer. Two issues:

1. **Eject crash**: `end()` teardown race on dual-core S3 — now fixed with
   drain-before-teardown.
2. **Cable disconnect freeze**: `tud_disconnect()` called when VBUS is already
   absent faults the S3 USB controller — now guarded.

## Root Cause

On the ESP32-S3 (dual-core Xtensa LX7), `UsbMassStorage::end()` tore down
TinyUSB MSC callback state in the wrong order, AND unconditionally called
`tud_disconnect()` — which faults the USB peripheral when the cable is
already physically disconnected (VBUS absent).

## Fix

1. Guard `tud_disconnect()` + drain loop in `if (tud_mounted())` — skip
   when the host is already gone.
2. Drain before teardown: `_active = false` → clear `gOwner`/`gDev` →
   `gMsc.end()` last.

## Verification

- Build (x4pro): passed
- clang-format: passed

### DIFF
--- a/libs/hardware/UsbMassStorage/src/UsbMassStorage.cpp
+++ b/libs/hardware/UsbMassStorage/src/UsbMassStorage.cpp
@@ -156,10 +156,24 @@ bool UsbMassStorage::begin(FsBlockDeviceInterface* dev) {
 
 void UsbMassStorage::end() {
   if (!_active) return;
-  gMsc.end();
+  // Signal the host to release the device BEFORE tearing down callback state.
+  // tud_disconnect() disables the D+/D- pull-up so the host ejects the media;
+  // we then wait for tud_mounted() to go false so no in-flight MSC callback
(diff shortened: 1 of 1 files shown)


Belphemur:
Just to be clear, fully tested and working on X4Pro

itsthisjustin:
> Just to be clear, fully tested and working on X4Pro

I haven't seen a crash on the pro when disconnecting. I believe it soft reboots on purpose though?

Belphemur:
On my side it would just freeze, not soft reboot.

No reaction to button press and staying on the normal mounted screen.

Those changes fixes it by showing disconnecting and returning to the home page.

-------- Original Message --------

PR #89, opened 2026-09-08
feat: expose grayscale base refresh support

uxjulia:
## Summary

- Adds `supportsAsyncGrayscaleBase()` as a way for each display driver to indicate whether it supports using a normal async refresh before grayscale is applied.
- Enables this by default for SSD1677 displays.
- Keeps it disabled for X3's UC8253 and UC8279d display types, since they need their own grayscale refresh sequence.

## Additional Context

This moves the display-specific decision into the SDK instead of having CrossPoint check whether the device is an X3.

Both current X3 displays need the special grayscale path, but keeping that detail with each display driver makes it easier to support future display variants correctly.

This does not change display behavior on its own. It only gives apps a safer way to choose the correct refresh path.

Relates to Crosspoint PR #3439 (which will be dependent on this) and issue #3400

### DIFF
--- a/libs/display/FreeInkDisplay/include/FreeInkDisplay.h
+++ b/libs/display/FreeInkDisplay/include/FreeInkDisplay.h
@@ -94,8 +94,10 @@ class FreeInkDisplay {
 
   // Frame buffer operations
   void clearScreen(uint8_t color = 0xFF) const;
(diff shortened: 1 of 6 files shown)


itsthisjustin:
I could swear I had this exact thing in the SDK already. Maybe I bailed on it?

uxjulia:
> I could swear I had this exact thing in the SDK already. Maybe I bailed on it?

There are definitely a bunch of similar looking APIs, but nothing matching this one as far as I can tell

PR #90, opened 2026-09-08
feat(keyboard): add Vietnamese keyboard layout

brianhuster:
Vietnamese has a lot of diacritics, so I believe it deserves is own
keyboard layout.

Vietnamese layout is different from English layout in 2 aspects:
- Diacritics row: this row is used to type letters with diacritics. The
  order is inspired by the VNI input method, which is the second-most
  popular Vietnamese input method pop
- No number row when unless `?123` is pressed. Diacritics row already
  make the layout tall enough, and users can still type numbers and
  punctuations by pressing `?123` button.

Why not TELEX (which is the most popular Vietnamese input
method)? Because it is very different from any existing input methods in
Crosspoint and will require more complex logic handling (for example,
typing `aa` must produce `â`, but typing `aaa` must produce `aa`)

### DIFF
--- a/libs/ui/FreeInkUI/include/components/keyboard/keyboard.h
+++ b/libs/ui/FreeInkUI/include/components/keyboard/keyboard.h
@@ -38,6 +38,9 @@ enum class KeyboardLayoutId : uint8_t {
   // Hebrew: no letter case, so a single layer and no shift key. Right-to-left
   // is the renderer's job -- the layout inserts code points in logical order.
   HebrewIl,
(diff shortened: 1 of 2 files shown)


itsthisjustin:
@brianhuster lots of keyboard changes coming soon fyi but I notice this is still in draft anyway. Just bumping for you to know

brianhuster:
I'm curious what the changes are about.

Btw the reason I set this as draft is because although the keyboard layout in this PR is optimized for non-touch devices, people who use touch devices would prefer an IME like the one they use on smartphones. However the IMEs on smartphones are much more complicated to implement.

PR #95, opened 2026-09-09
fix(uc8279): vertical banding on dithered gray images on the X3

bos:
## Summary

Adds an absolute four-tone image grayscale path to the UC8279 Xteink X3 driver, using the stock XTH4 waveform that was already captured in `Uc8279X3Luts.h` but unused.

- `PanelDriver::supportsAbsoluteGrayPlanes()` (default false) and its `FreeInkDisplay` forward declare that `displayGray(factoryMode=true)` runs an absolute four-tone waveform and consumes absolute planes: LSB plane bit = white or dark gray, MSB plane bit = white or light gray, so both set = white and both clear = black. Plain black-and-white content is just the B/W frame in both planes. Non-factory passes keep the differential mask encoding.
- `Uc8279Driver` returns true and, for `factoryMode`, loads the stock XTH4 tables with the register assignment that maps those plane pairs onto the rows that settle to black, dark gray, light gray and white. The anti-aliasing path is unchanged.

Stacked on #94 (the dark-gray register exchange); only the second commit is new here.

## Why

The X3's short AA nudge exposes an 8-gate-line drive nonuniformity on any dithered gray image: light-gray pixels next to differently driven neighbours come out with a column-periodic brightness variation of about 4% of the black-to-white range, visible as vertical stripes. Flat fields and monochrome dither are clean, and reshaping or lengthening the nudge (10 and 14 frame variants, split and interleaved pulses) did not change the period or the amplitude. The long, DC-balanced XTH4 waveform does.

The X4 Pro driver already takes this approach (e85297e, scaled XTH4 for image-classified passes). The X3 has no PSRAM for the base-frame fold that implementation uses, so here the host supplies absolute planes directly; CrossPoint can do that because its image paths re-decode the source per plane pass. A CrossPoint PR that uses the capability follows.

## Validation

Xteink X3, UC8279 (ZHX368). Test image with flat patches at all four levels, a light-gray/black checkerboard and line patterns, photographed in raw and measured relative to the black and white patches.

| | AA nudge (shipped) | XTH4 absolute pass |
| --- | --- | --- |
| checkerboard column banding, p2p % of range | 3.5 – 4.7 | 0.9 – 1.1 |
| dominant column period | 8 gate lines | none at 8 |
| dark gray position | 0.21 – 0.24 (with #94) | 0.17 – 0.27 |
| light gray position | 0.77 – 0.79 | 0.67 – 0.69 |
| refresh time | ~0.1 s | ~1 s |

Levels come out in the right order with the unscaled stock table. Dark gray is darker than ideal and can be tuned later with a time-scaled bank as on the X4 Pro; that is left out of this PR so the validated stock bytes ship first.

## Notes

- The waveform is not fully absolute from a white start: an area drawn white in the base but left at (0,0) in the planes settles dark gray rather than black. Hosts therefore draw overlays into the planes as well; the CrossPoint change does.
- The photos still show an irregular vertical variation of about 1% of range at a scale of roughly 20 columns. Its position tracks where the panel sits in the camera frame between shots rather than the panel itself, so it is most likely the camera's own column pattern; the panel residual, if any, is below what these measurements resolve.

### DIFF
--- a/libs/display/FreeInkDisplay/src/driver/Uc8279Driver.cpp
+++ b/libs/display/FreeInkDisplay/src/driver/Uc8279Driver.cpp
@@ -17,7 +17,11 @@ constexpr uint8_t CMD_DTM1 = 0x10;                // OLD plane in KW mode
 constexpr uint8_t CMD_DATA_STOP = 0x11;           // DSP
 constexpr uint8_t CMD_DISPLAY_REFRESH = 0x12;     // DRF
 constexpr uint8_t CMD_DTM2 = 0x13;                // NEW plane in KW mode
-constexpr uint8_t CMD_LUT_VCOM = 0x20;             // first LUT register (0x20-0x24)
+constexpr uint8_t CMD_LUT_VCOM = 0x20;            // LUT registers 0x20-0x24
+constexpr uint8_t CMD_LUT_WW = 0x21;
+constexpr uint8_t CMD_LUT_BW = 0x22;
+constexpr uint8_t CMD_LUT_WB = 0x23;
+constexpr uint8_t CMD_LUT_BB = 0x24;
 constexpr uint8_t CMD_VCOM_DATA_INTERVAL = 0x50;  // CDI
 constexpr uint8_t CMD_PARTIAL_WINDOW = 0x90;      // PTL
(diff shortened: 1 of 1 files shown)


itsthisjustin:
I manually merged this into existing work. Thanks so much! You're credited in the commits

PR #97, opened 2026-09-13
fix(ssd1677): keep the wake-clean promotion when turnOff is set

eszter007:
### What goes wrong

The following issue occured in my crosspoint fork matcha reader with sunlight fading fix on certain image pages: https://github.com/eszter007/matcha-reader/issues/237
When the user puts a device to sleep, the sleep image stays on screen if sunlight fading fix is on. By that, the page looks frozen as the new page has already been drawn into the framebuffer, so the device may appear frozen.
 Turning a page usually clears it, and then everything behaves normally again.

It only happens when the firmware driving this SDK asks for the panel to be powered down at the end of every refresh, which is the `turnOff` argument on `display()`.

The fix in the SDK enables the fixes for all paths instead of needing forks like mine to overwrite it only for specific views for which bugs are reported. The SDK fix thus enables more stability and a root improvement.

The following is AI reasoning (Claude Opus 5), which was verified on my physical X4 (the device where the bug was reported):

### Where that `turnOff` comes from

In my case it comes from a CrossPoint setting called the "sunlight fading fix". Some panels wash out and look faded when the device is read in direct sunlight with the panel's power left on between page turns, and this setting exists to avoid that. All it does is ask for the panel to be powered down after each refresh, so CrossPoint passes `turnOff=true` on every `display()` call for as long as the setting is on.

There is nothing unusual about it beyond that. It is a normal consumer of this driver that happens to pass a parameter on every call, where most callers only pass it now and then.

Worth saying because it made the bug hard to place: the setting is an exact on/off switch over the failure. With it on, every wake leaves the old screen up. With it off, wake works perfectly. A checkbox in the reader's display settings appearing to control whether the screen redraws points at the app, so that is where I spent a while looking before this turned out to be in the driver.

### Why the driver ends up in that state

Some background on the two refresh kinds involved. A FAST refresh is differential: the controller keeps the previously displayed image in its RED RAM plane, and FAST only drives the pixels that differ from it. That is what makes it quick. A HALF refresh ignores that plane and drives every pixel to its target value, so it can always clear the screen no matter what is on it.

Deep sleep discards controller RAM. After a wake, the RED plane no longer holds the previous image, so a FAST refresh is comparing the new frame against nothing meaningful and leaves most of the panel untouched. Whatever was physically on the glass, in this case the sleep image, simply stays.

The driver already knows this. `initController()` arms a flag, `_needsInitialFull`, whose entire job is to force the first refresh after boot or wake to be a clearing one instead of a differential FAST. The comment above it names the sleep screen as the exact thing it is there to prevent.

The problem is that the flag is checked inside a block gated on `!turnOff`:

```cpp
if (!turnOff) {
  if (_needsInitialFull) {
    if (mode == RefreshMode::Fast) mode = ...Half/Full;   // promote
    _needsInitialFull = false;                            // consume
  } else if (!_isScreenOn && _cfg.fullSeqOverride == 0) {
    mode = RefreshMode::Half;
  }
}
```

A caller that always powers the panel down never enters that block. The flag is therefore never acted on and never cleared. It stays armed for the entire session while the first paint after every wake goes out as a plain FAST.

This affects boards that rely on `_needsInitialFull`, meaning those configured with a vendor refresh sequence (`fullSeqOverride`). The X4 is one of them.

### The change

The `_needsInitialFull` check now runs whether or not `turnOff` was requested. Which waveform a paint needs does not depend on whether the panel's power drops afterwards. The flag is armed once per `begin()` and cleared the first time it fires, so the cost is at most one promoted refresh per boot.

I deliberately left the second branch gated. It reads `!_isScreenOn` as a sign that the panel is cold and needs a clearing refresh, which is a reasonable signal for an ordinary caller. For a `turnOff` caller it says nothing at all, because the panel is off before every single paint for the simple reason that this caller just turned it off. Removing the gate there would make every page turn a slow, visibly flashing full refresh.

### How it was verified

`test_pro.cpp` gains a case next to the existing `testSsd` promotion check, same shape but passing `turnOff=true`. After `begin()`, the first FAST request must activate `0xd7` (the clearing HALF sequence) and the second must activate `0xfc` (the differential FAST sequence), showing that the one-shot both fires and is consumed.

It checks whether a value was written at all rather than reading the last write, because `0xfc` does not power the panel off by itself, so the driver follows it with a separate `0x22=0x03` power-down. The last write to `0x22` is that power-down, not the waveform.

The test fails on `main` and passes with the fix. All four host suites pass.

Also confirmed on hardware (X4, SSD1677) with `SSD1677_PROBE_DEBUG=1`. Waking onto a page, before the fix:

```
[810]  [ACT] Entering activity: EpubReader
[EPD]  displayBuffer mode=2 off=1
[SSD1677] FAST refresh 502ms (ctrl2=0xfc, seq)     <-- first paint after wake
```

`off=1` is `turnOff`, so the fading fix was on, and `ctrl2=0xfc` is the differential FAST that cannot clear the panel. With the fix the same paint activates `0xd7`, the screen clears, and the page is there on wake.


itsthisjustin:
Do we also need this for x3s which also have the sunlight fading fix?

eszter007:
> Do we also need this for x3s which also have the sunlight fading fix?

I checked that quickly with AI since I don't have a X3 myself, by just letting it check the X3 code and it came back with the following reasoning (after it had written a test just for the X3 which was a bit unnecessary to commit since it was just to prove the point)^^:

The X3 doesn't need it. It does have the same-looking gate, in `Uc8253X3Driver::displayStart`:

```cpp
if (!_isScreenOn && !turnOff) {
  mode = RefreshMode::Half;  // wake transition gets a stronger waveform
}
```

The difference is what that gate controls. On the X3 it only relabels the mode. The clean itself comes from `doFullSync` a few lines below, which is driven by `_initialFullSyncsRemaining` and `!_redRamSynced` — `begin()` sets both — and never looks at `turnOff`. So the first paint after a wake takes the full-sync branch whether or not the caller asked for a power-down, and the sleep screen gets cleared either way.

The SSD1677 has no equivalent. Its one-shot lives *inside* the gated block, so a `turnOff` caller never reaches it, never consumes it, and the flag just sits armed for the whole session while every wake paints differentially over a RAM that deep sleep discarded.

Worth adding that the X3 gate is load-bearing, so removing it for symmetry would make things worse rather than better. A caller passing `turnOff` leaves `_isScreenOn` false before every single paint, so promoting on that alone would turn every page turn into a half scrub. That's the same reason I left the `!_isScreenOn` branch gated on the SSD1677 side and only ungated the one-shot.

I confirmed the first half against the host harness locally: with `turnOff` set, the first paint after `begin()` takes the full-sync branch and the paint after it stays differential.

PR #101, opened 2026-09-15
Panel_EPD: make a clean refresh scrub even when it follows another one

jpirnay:
## Summary

On a board whose clean LUT rail-normalizes before it lands, a clean refresh that follows **another** clean refresh leaves the previous page on the glass as a faint imprint. This stops that, without changing which bank any refresh uses and without a second waveform.

Opt-in per board; no board in this tree changes behaviour.

> **Rewritten.** The first version of this PR downgraded `Half` to the differential bank in the back-to-back case. That was wrong in both directions — the fallback fired exactly where a scrub was wanted, and the case it was avoiding turns out to cost **187 ms**, not a waveform. Device testing caught it. Details at the bottom.

### The mechanism

`Panel_EPD`'s `epd_text` branch drives a pixel unless it was **already requested white under that same bank** and is requested white again — the `white != d1 || d1 != s0` test, in which `white` embeds the bank's own LUT offset:

```cpp
uint_fast16_t white = lut_offset | 0x00FF;
...
if (white != d1 || d1 != s0) { /* drive */ }
```

Because the offset is part of the comparison, **the bank a pixel was last driven under is part of its identity**:

| previous refresh | what a clean push drives |
| --- | --- |
| any other bank | every pixel — offsets differ, so nothing compares equal. A proper scrub. |
| the clean bank | only the union of old and new ink. The white background is skipped. |

On a table that drives every touched pixel to black, then to white, then down to its level, the second row reads as **both pages standing at once** before the new one resolves — and the old ink then settles beside a background that was never driven, leaving its shape behind.

That is not a rare sequence. `Half` is exactly the mode a host spends on the transitions that *should* scrub: one armed to enter a reader, another to come back from it, and a screen-repair cycle is a dozen of them back to back.

### The fix: stop the skip, don't change bank

Nothing has to reach the glass to clear that condition. Only the offset stored in the reserved half of `_step_framebuf` has to differ, and one full-screen write through **any other bank** changes it for every pixel at once.

`blit_dmabuf` supplies the instrument:

```cpp
if (s_0 >= 0) {
  tmp = lut[s_0];
  s_0 += 256;
  buf += tmp << 4;
  if (tmp == 0) { s_0 = src[1]; tmp = s_0 | 0x8000; src[1] = tmp; }
  src[0] = s_0;
}
```

A **zero LUT entry means "this pixel is finished"** — load the reserved request, mark idle, drive nothing. So a bank that is a single `0u` word is a complete refresh that ends after one frame and never touches the ink, *while still leaving its own offset behind*. Spare LUT slots commonly already hold exactly that, so the driver derives availability from the config rather than taking another flag.

It must be `epd_fastest`, **not** `epd_quality`, even though both slots tend to be spare: the fast modes take the `flg_fast` branch, which writes `d[0] = s0 - 0x8000` and goes straight to the bank, while the others prepend `lut_eraser` — the pixel would run the eraser's two-frame nudge and then stop with nothing driving it home. That is a half-inverted screen that never resolves, and it is exactly what this panel showed when `Full` was once routed to `epd_quality`.

A board with no such slot drives the white rail through the differential bank instead: a real refresh, so a real cost, but known-good (it is the drive every page turn makes) and on the glass it reads as the white flash a scrub is expected to open with.

## Testing

Measured on a **LilyGo T5 S3** supplying its own ED047TC2 waveform, at 24 °C (waveform range 3), with `LGFX_EPD_PUSH_TRACE`:

```
HALF 2579 ms  sprite=115 settle=2414   no re-tag (follows the differential bank)
HALF 2766 ms  sprite=115 settle=2409   [epd] normalize: re-tag through fastest
```

`sprite` and `settle` are **identical**, so the whole **187 ms** delta is the re-tag, sitting outside the traced push. The white-flash path would have cost 35 blocks × 35.4 ms/frame = **1239 ms** — so the no-drive bank is **6.6× cheaper**, and invisible on the glass.

A full trace of a screen-repair cycle (thirteen consecutive `Half` pushes, the worst case for this) shows all thirteen landing on `text(clean bank, eraser)` with a `fastest` re-tag on each one after the first.

**Device-confirmed by a user:** reader entry scrubs again, no ghosting on reader exit, and no ghosting buildup after the repair cycle.

**Unchanged for every board in this tree.** `cleanBankNeedsFreshBackground` defaults false, which keeps `epdModeFor()` returning `epd_text` for both `Half` and `Full` and skips the normalizing pass entirely — so `M5PaperS3`, the `LgfxEpd` board here, is byte-for-byte as before.

## What changed since the first version of this PR

The original proposed resolving `Half` against the previous refresh and falling back to the differential bank when the clean one could not scrub. On hardware that produced ghosting in two places — returning from a reader, and every screen after a repair cycle — because the fallback delivers a **full-screen DU** where a scrub was expected. It also mishandled the repair cycle worst, which is the one place the "back-to-back" case is guaranteed.

The replacement never downgrades. Since the re-tag measures 187 ms rather than the ~1.2 s a real refresh costs, there was never a reason to trade the scrub away.

## Note for reviewers

This and #102 both insert into `LgfxEpdDriver::begin()` just after `setEpdMode(epd_fast)`, so they will conflict textually whichever merges second. The additions are independent and the resolution is trivial; I will push a rebase for whichever lands second.

## AI usage

PARTIALLY. An AI coding assistant was used for the `Panel_EPD` analysis, for drafting the code comments and this description. The hardware measurements and the device validation were done by hand on the board.


jpirnay:
Cross-reference: #102 touches `LgfxEpdDriver::begin()` at the same anchor as this one (both insert just after `setEpdMode(epd_fast)`), so the two will conflict textually whichever merges second. The additions are independent — that one only prints the LUT block total, this one changes refresh routing behind an opt-in flag — so the resolution is trivial and I will push a rebase for whichever lands second.

jpirnay:
Rewritten — the previous version of this PR was wrong, and device testing is what caught it.

**What it used to do:** resolve `Half` against the previous refresh and fall back to the differential bank when the clean one could not scrub, on the theory that the back-to-back case was rare.

**Why that was wrong:** the fallback delivers a *full-screen DU* precisely where a scrub was expected. On hardware that produced ghosting returning from the reader, and on every screen following a screen-repair cycle — which is thirteen consecutive `Half` pushes, so the case I had assumed was rare is guaranteed there.

**What replaced it:** never downgrade. Instead clear `epd_text`'s skip condition by re-tagging the background through a no-drive bank — one frame, no ink driven, and it leaves the LUT offset the clean bank needs to see. Measured at **187 ms** against **1239 ms** for a real differential refresh, so there was never a reason to trade the scrub away in the first place.

The trace that settled it, at 24 °C on an ED047TC2:

```
HALF 2579 ms  sprite=115 settle=2414   no re-tag (follows the differential bank)
HALF 2766 ms  sprite=115 settle=2409   [epd] normalize: re-tag through fastest
```

`sprite` and `settle` identical — the whole delta is the re-tag, outside the push.

Device-confirmed on a LilyGo T5 S3: reader entry scrubs, no ghosting on reader exit, no ghosting buildup after the repair cycle. Boards leaving the flag false are byte-for-byte unchanged.

Apologies for the churn on a PR you may already have started reading.

PR #102, opened 2026-09-15
Panel_EPD: report the LUT block budget at init, because overrunning it fails silently

jpirnay:
## Summary

A `Panel_EPD` board whose LUT banks total 128 blocks or more fails **silently and globally**, and nothing in the stack says so. This computes the total at init and prints it.

Diagnostics only — no behaviour change.

### The limit, and why it is invisible

`Panel_EPD` packs a pixel's refresh progress into a `uint16_t` as `(lut_block << 8) | level`, and `blit_dmabuf` reads it back through a **signed** cast (`l16si` in the Xtensa path, `(int16_t)` in the C fallback) and skips the pixel when the result is negative. Bit 15 therefore means *this pixel is idle* — so no block index may ever reach 128, and the five banks together have to fit in 128 blocks whichever bank a given refresh uses.

Overrun it and every pixel whose waveform reaches block 128 reads as idle mid-refresh. Refreshes stop completing and the panel simply looks dead. There is no error, no assert, and no log line.

Worse, it does not have to be reproducible: a board selecting a waveform per temperature generally needs **longer** banks at the cold end, so the same firmware can work on a warm bench and blank in a cold room. That is exactly how we met it, on a board supplying its own tables, and it cost a debugging session that started in entirely the wrong place.

### What this adds

`lgfxEpdLutBlocks()` mirrors the accumulation in `Panel_EPD::init_intenal()`, including:

- the hardcoded 4-block `lut_eraser` bank prepended to every `epd_text` / `epd_quality` refresh, which config never supplies and is easy to forget;
- the stock table `Panel_EPD` substitutes into any slot a config leaves empty (quality 32, text 32, fast 10, fastest 7 — a board on all four already pays 85 of the 128).

`begin()` then prints either a loud over-budget message naming the failure mode, or the headroom. The headroom is the number that matters: a board regenerating its own tables needs it at the moment it changes them, not after the panel goes dark.

```
[epd] LUT budget 105/128 blocks (23 spare), 53760 B internal DMA of which 26880 B is upstream slack
```

### The byte figure

It rides along because `Panel_EPD` sizes `_lut_2pixel` as `blocks * 256 * sizeof(uint16_t)` but indexes it as bytes (`l8ui`), so **half of that internal DMA allocation is never touched** — 26 KB at the long end of a per-temperature waveform, out of internal RAM that a `Panel_EPD` board tends to be short of. Not fixable from here (private member, no accessor), but worth being able to see when internal RAM is tight, and it is free to print next to a number already being computed.

## Testing

Measured on a LilyGo T5 S3 at 24 °C, where it reports `105/128 blocks (23 spare)` — matching a hand count of the emitted banks, terminators included. The over-budget branch was confirmed by the failure it was written for: an earlier waveform on that board totalled 143 blocks at its coldest range and blanked the display below roughly 27 °C.

Builds for the `LgfxEpd` envs. No functional paths touched.

## AI usage

PARTIALLY. An AI coding assistant was used for the `Panel_EPD` analysis, for drafting the code comments and this description. The measurements were taken on the board.


jpirnay:
Note for whoever reviews these together: this and #101 both insert into `LgfxEpdDriver::begin()` immediately after `setEpdMode(epd_fast)`, so they will conflict textually whichever lands second — trivially, since the two additions are independent and neither reads the other's state. Happy to rebase this one on top of #101, or the other way round; just say which you would rather take first and I will push the rebase same-day.

They are otherwise unrelated: #101 changes refresh behaviour behind an opt-in flag, this one only prints a number.

PR #104, opened 2026-09-16
fix: recover from an aborted deep-sleep entry instead of spinning

Chuckthe5th:
## What

`PowerManager::deepSleep()` calls `esp_deep_sleep_start()`, which is documented
as not returning. The fallback for the case where it does return anyway was an
infinite busy-loop:

```cpp
esp_deep_sleep_start();
while (true) {
}  // esp_deep_sleep_start() does not return; satisfy [[noreturn]]
```

That loop draws full-clock current (~20-30 mA) forever and never responds to
the power button. On the outside, that's indistinguishable from a real
sleep-current drain bug -- which is exactly what made this the wrong fallback
while chasing one (see #x3-sd-rail-power-gating investigation, where I first
had to rule out an aborted-sleep loop as the cause before finding the actual
X3 GPIO13/`sd.powerEnable` gap, already merged in fde240fa).

## What changed

Replace the busy-loop with `esp_restart()`. The abort is recorded first in an
`RTC_NOINIT_ATTR` struct (survives the warm reset that `esp_restart()`
triggers, magic-guarded so a cold boot doesn't read garbage), so the next
boot's consumer code can check `PowerManager::takeAbortedSleepInfo()` and
report/log it instead of the failure silently presenting as a drained
battery.

## Scope note

This PR is intentionally narrow. It was split out of a broader local fix
after confirming the X3 SD-rail piece (`XTEINK_X3.sd.powerEnable`) is already
merged upstream (fde240fa), and that `SDCardManager`'s SPI-backend
`shutdown()` is an intentional no-op by design (per its own comment: cutting
the rail is considered sufficient) -- so neither belongs in this PR. Only the
sleep-abort-recovery fix is novel against current `main`.

## Testing

Built clean against the ESP32-C3 (Xteink X3) target. The abort path itself is
not easily reproducible on real hardware (it requires the SoC to reject sleep
entry, which is rare) -- verified by code inspection and that the surrounding
`deepSleep()`/`deepSleepUntilPowerButton()` control flow is otherwise
unchanged.


itsthisjustin:
Have not seen this in real life yet. How are you triggering it?

Chuckthe5th:
> Have not seen this in real life yet. How are you triggering it?

I'm not, and I haven't actually seen it fire either.

I was chasing a sleep drain on the xteink X3. Turned out to be the
sd.powerEnable gap (fde240f), card rail staying live through sleep. While
narrowing it down I stuck a log line after esp_deep_sleep_start() to rule out
"sleep entry is getting rejected" as a cause. Never fired, across about 200
hours of sleep testing. But that's when I noticed the fallback was an
infinite busy-loop.

I submitted this PR because if it ever _does_ hit, a spin at full clock is ~20-30 mA
and won't respond to the power button. Device just flattens its battery and
looks exactly like a drain bug. Took me a while to rule that out specifically
because I couldn't tell the two apart from the outside. esp_restart() costs
nothing, satisfies [[noreturn]] the same way, and the RTC_NOINIT record lets
the next boot say what happened.

PR #109, opened 2026-09-18
fix(display): keep the displayed frame when leaving dual-buffer mode

jpirnay:
## The bug

In dual-buffer mode `displayBuffer()` ends in `swapBuffers()`, so:

- `frameBufferActive` holds the frame **on the panel**;
- `frameBuffer`, the write target, holds the frame from **two refreshes ago**.

Both doors out of dual-buffer mode drop the secondary and stop there:

- `releaseSecondaryBuffer()` frees whichever slot `frameBufferActive` points at and nulls it — that is, it frees *the displayed frame*;
- `borrowSecondaryBuffer()` hands those same bytes to the host as scratch and nulls the pointer ("single-buffer mode, same as `releaseSecondaryBuffer()`").

Either way the write buffer is left holding the two-generations-old frame while the display is now in single-buffer mode, where every other path assumes the write buffer *is* the on-screen frame.

`syncWriteBufferFromActive()` cannot repair it afterwards, because its memcpy is gated on the pointer that was just nulled:

```cpp
if (frameBuffer && frameBufferActive) memcpy(frameBuffer, frameBufferActive, bufferSize);
```

It returns having done nothing, with no error.

## Why it has gone unseen

A full render clears first, so it never notices. Only a **partial** repaint reads the write buffer, and every partial repaint I could find runs inside the same activity that released the buffer. The first code to composite over a frame a *different* activity drew shows it immediately: a pull-down control-centre sheet (`sheet` + `sliderRow`) opened over the reader paints onto the **previous** screen. The activities that free the secondary for headroom are ordinary ones — a cover loader, a chapter build.

## The fix

Carry the displayed frame into the write buffer before the secondary goes away, at both doors.

That makes true an invariant `reallocSecondaryBuffer()` already states in its own comment —

> correct when the caller reallocs before drawing the next page (frameBuffer then still holds the on-screen frame)

— which was false immediately after a release. `returnSecondaryBuffer()` re-seeds from the same place, so the borrow path needs it for the same reason, and there the bytes are about to be scribbled on by the host, so it is the last chance to keep them.

Cost is one framebuffer-sized memcpy (~48 KB) on a path that is about to spend hundreds of milliseconds parsing CSS or decoding an image.

## Testing

Device-confirmed on a LilyGo T5S3: the drawer paints on the correct page from the home screen, from the reader, and from the reader with a background chapter build in flight. The equivalent carry has been running in our HAL wrapper around `releaseSecondaryBuffer()` since; this PR moves it to where the invariant lives, and adds the borrow path the wrapper could not reach. Firmware builds green on ESP32-C3 with the change in the SDK.

## Related, deliberately not changed here

`releaseSecondaryBuffer()` frees without draining a refresh in flight — `borrowSecondaryBuffer()` calls `syncPendingAsync()` first, the release path does not. On a panel whose post-waveform sync re-reads the previous frame (X3's DTM1 pass) that means freeing memory the driver still reads, so a caller has to order its release against any async refresh by hand. Left alone to keep this PR to one concern; happy to send it separately if you want it fixed here.


itsthisjustin:
48kb is really big. Is that better than the issue here?

PR #110, opened 2026-09-18
fix: render keyboard delete icon at native size

uxjulia:
## Summary

- Replace the 16px keyboard delete icon with a 28px mask generated from Lucide’s `delete.svg`.
- Render keyboard glyphs at their native size, capped by the available key dimensions.
- Use the improved delete icon in both the full keyboard and key-grid components.
- Add a host test confirming the QWERTY delete icon renders at 28×28.

## Why

The previous 16px icon was enlarged using nearest-neighbor scaling. This produced uneven, pixelated strokes on larger keyboard keys.

Generating the icon at its intended display size keeps the outline crisp without adding runtime SVG rendering or heap allocations. The larger static mask adds approximately 80 bytes of flash.

## Validation

- FreeInkUI host tests: 225,939 checks passed
- Regenerated the asset from the pinned Lucide source and confirmed it matches the embedded mask

---

**Crisp delete icon now:**

delete key not pixelated


itsthisjustin:
Thanks I noticed this too and needed to fix it

